PR #1, opened 2025-12-16
Bump of nvidia/cuda version to 13.1.0

ateska:
I don't know if you accept PRs ... but here it is.

Tested on my dual NVidia DXG Spark stack.


eugr:
I'll test on my end and merge

PR #3, opened 2025-12-18
add dependenies to allow plot output from vllm bench

christopherowen:


eugr:
how much does it add to the container? I removed it from the base build because it was pulling the entire x.org with it. But I was using regular gnuplot, not gnuplot-nox

eugr:
Yeah, it pulls 234MB into the container. Better than a regular gnuplot that pulls 500MB, but still. Maybe introduce an argument to control this (off by default)?

eugr:
I'll add a parameter to the build script to install these dependencies in the runtime image if needed (off by default).

eugr:
Since we are now using llama-benchy, I'm closing this

PR #4, opened 2025-12-18
Improve rebuild times by caching encodings

christopherowen:


eugr:
It's just 5 megs though - download would be the same speed on any decent connection? Maybe we should also include datasets for vllm bench serve here, so we don't waste the cache?

PR #6, opened 2025-12-18
Limit build parallelism to reduce OOM situations

christopherowen:


eugr:
Any reason not to set it to 20 by default, so we could get the maximum build speed?

christopherowen:
20 should be the current default without setting anything, I believe.

I was experiencing unresponsiveness from the spark while compiling (I was doing other things over ssh and have other services running on it).

I found that 15-16 was a sweet spot to keep it responsive and not OOM - which happened to me a few times on wiped cache full builds.

eugr:
OK, I'll merge then and add an additional argument to build script.

PR #10, opened 2026-01-16
Add some utility methods to support revert mods and download mods

raphaelamorim:
This PR introduces a comprehensive system for managing modifications (patches) to the vLLM container environment. It adds functionality to apply, revert, list, and automatically download mods from GitHub Pull Requests, streamlining the process of testing fixes and custom configurations.

Key Changes: 

> Mod Management in launch-cluster.sh:

1. Added --revert-mod <path> to undo applied patches.
2. Added --list-mods to display available mods with dynamically wrapped descriptions.
3. Added --download-mod <PR_URL> to automatically fetch a GitHub PR, sanitize the diff for the container environment, and generate a compatible run.sh script.

> Standardized Mod Interface:

Refactored run.sh structure to support apply(), revert(), and description() functions.
Updates to launch-cluster.sh to call these specific functions rather than just executing the script.

> Path Convenience:

Automatic resolution of paths in the [mods](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) directory (e.g., --apply-mod my-mod now resolves to mods/my-mod).

> Documentation:

Updated README.md with a new "Mod Management" section detailing usage examples for creating, listing, and downloading mods.

> Support sequential operations and mixed actions in a single command.

I changed MOD_PATHS to MOD_QUEUE_PATHS (and introduced MOD_QUEUE_ACTIONS and MOD_QUEUE_TYPES) to support sequential operations and mixed actions in a single command.

Previously, if we just had a list of paths, we wouldn't know if the user wanted to apply or revert them, or in what order.

By using parallel arrays:

MOD_QUEUE_PATHS: Stores the path to the mod.
MOD_QUEUE_ACTIONS: Stores "apply" or "revert".
MOD_QUEUE_TYPES: Stores "dir" or "zip".

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -24,9 +24,8 @@ DAEMON_MODE="false"
 CHECK_CONFIG="false"
 ACTION="start"
 CLUSTER_WAS_RUNNING="false"
-MOD_QUEUE_PATHS=()
-MOD_QUEUE_TYPES=()
-MOD_QUEUE_ACTIONS=()
+MOD_PATHS=()
+MOD_TYPES=()
 
 # Function to print usage
 usage() {
@@ -38,7 +37,6 @@ usage() {
     echo "  --ib-if         InfiniBand interface (Optional, auto-detected)"
     echo "  --nccl-debug    NCCL debug level (Optional, one of: VERSION, WARN, INFO, TRACE). If no level is provided, defaults to INFO."
     echo "  --apply-mod     Path to directory or zip file containing run.sh to apply."
-    echo "  --revert-mod    Path to directory or zip file containing run.sh to revert."
     echo "  --download-mod  Download a GitHub PR as a mod (e.g., https://github.com/vllm-project/vllm/pull/31386)"
     echo "  --list-mods     List available mods in mods/ directory."
     echo "  --check-config  Check configuration and auto-detection without launching"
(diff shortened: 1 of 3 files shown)


launch-cluster.sh
```diff
@@ -52,7 +97,138 @@ while [[ "$#" -gt 0 ]]; do
...
+                mkdir -p "$mod_dir"
+                echo "Downloading patch from $diff_url..."
+                if ! curl -sL "$diff_url" -o "$patch_file"; then
+                    echo "Error: Failed to download patch."
+                    rm -rf "$mod_dir"
+                    exit 1
+                fi
+
+                # Clean up git headers and adjust paths to match container python location
+                # This ensures we can apply the patch from root (/) just like other mods
```
eugr: I think, just specifying `-d /usr/local/lib/python3.12/dist-packages` in the patch command is a cleaner way to do it, since you generate `run.sh` anyway.

eugr:
Good idea regarding auto-generating mods from PRs, but I don't know if reverting a mod makes any sense. Since the containers are stateless, all changes that you make by applying the mods are gone after you restart the cluster. 

Moreover, unless I missed something, any changes you make AFTER the cluster is started (add, revert, doesn't matter), will not be applied as the process is already running and there are no live reloads. It can only introduce some extra bugs, unless you kill the process, but at that stage it would be easier to just restart the whole thing again (and guarantee that we don't leave any artifacts).

PR #14, opened 2026-01-19
01 readme update

corinabioinformatic:
Updated README, related to issue 13 ,https://github.com/eugr/spark-vllm-docker/issues/13

some other updates are the .gitignore and .dockerignore just in case


eugr:
Won't merge as that may confuse people who don't want to use this feature and prevent automatic backend selection for the majority of users.

PR #19, opened 2026-01-25
Adding Runtime Profile/Recipe System

raphaelamorim:
@eugr, I was thinking about Introducing a profile system that decouples runtime configuration from the launch-cluster.sh script. Instead of specifying long command lines with runtime-specific parameters, users can now define reusable JSON profiles.

Motivation
Previously, switching between runtimes (vLLM, SGLang, TensorRT-LLM) or model configurations required editing lengthy command lines:

```
./launch-cluster.sh exec vllm serve QuantTrio/MiniMax-M2-AWQ \
  --port 8000 --host 0.0.0.0 --gpu-memory-utilization 0.7 -tp 2 \
  --distributed-executor-backend ray --max-model-len 128000 \
  --load-format fastsafetensors --enable-auto-tool-choice \
  --tool-call-parser minimax_m2 --reasoning-parser minimax_m2_append_think
```
So, I've added a new --profile option in launch-cluster.sh to load runtime configuration from JSON files.

Usage:

```
# Before
./launch-cluster.sh exec vllm serve QuantTrio/MiniMax-M2-AWQ --port 8000 ...

# After
./launch-cluster.sh --profile example-vllm-minimax exec or 
./launch-cluster.sh --profile example-vllm-minimax
```

Dependencies
Requires jq for JSON parsing


dbotwinick:
@raphaelamorim 

Fun/interesting. I definitely like the idea since I'm actually doing something similar in what I plan to release eventually as open source that also includes a webUI to help manage hosts, setup clusters, manage & deploy inference "recipes" (there's the similar part). Basically I've been doing approximately the same thing but 1) image level (so what container image it's built on is part of the "recipe") and 2) all shell although I was considering if JSON would be a better interchange format. Was planning to also release that webUI tool as a docker container and then as x86+ARM so that it could run locally on spark or on user's desktop and then control spark via SSH. 

The reason I mention it here when it seems wildly unrelated is because I was thinking of trying to have a git repo that would be an exchange/repository for people's confirmed recipes to make it easier for people to get started. And if I go that route... and profiles are coming this way... it would be good if we more or less standardized representations so that it becomes easy to trend towards a unified ecosystem. 

One thought that would be relevant is if the image could/should be a field in the profile (with some sort of placeholder/default value for local build/image?). Obviously I am leaning that way, but it's a bit more nuanced here. But I like that direction because then if something worked in version XX, you can easily run it with matching version  [also why I am going around publishing versioned images...]

Other thoughts (feel free to reject... I'm just rambling):
runtime is currently equivalent of entrypoint in docker; command is then 2nd part which is sort of the first argument; might be easier if command is the executable/first thing and then args is just a text list and we entirely leave interpretation of args to the shell and command. In my WIP stuff, I separated args related to the runtime (VLLM_ARGS) and args related to the model (MODEL_ARGS) but that might be contrived/unnecessary. I would maybe add top level items for tp and ep even if it becomes redundant to the args because I foresee a future where people can browse profiles/recipes via a dashboard, so filtering by parallelism would be good. Even with what I said about command/runtime, I do like listing the runtime (again, thinking about future filtering). Other idea: add env field. configuration of some models/advanced options might require environment variables to be configured.

dbotwinick:
So e.g. in what I have been doing; here is bash script for glm-4.7-awq4-tp4, just to put something alongside my long, rambling comments, sorry about that...

```
# filename: glm47_awq4_tp4

#!/bin/bash

INFERENCE_SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"

VLLM_USE_DEEP_GEMM=0
VLLM_USE_FLASHINFER_MOE_FP16=1

IMAGE="scitrera/dgx-spark-vllm:0.14.0-t5"
MODEL="QuantTrio/GLM-4.7-AWQ"
VLLM_OPTS=(
  "-tp" "4"
  "--distributed-executor-backend" "ray"
  "--gpu-memory-utilization" "0.75"
  "--attention-backend" "FLASHINFER"
)
MODEL_OPTS=(
  "--trust-remote-code"
  "--tool-call-parser" "glm47"
  "--reasoning-parser" "glm45"
  "--enable-auto-tool-choice"
  "--served-model-name" "glm-4.7"
  "--speculative-config.method" "mtp"
  "--speculative-config.num_speculative_tokens" "1"
)
ENVIRONMENT_OPTS=(
  "VLLM_USE_DEEP_GEMM"
  "VLLM_USE_FLASHINFER_MOE_FP16"
)

source "$INFERENCE_SCRIPT_DIR/.vllm_template.sh"
```

and basically that generated docker command to create ray head+worker nodes and then launch that image... similar stuff overall...

raphaelamorim:
> @raphaelamorim
> 
> Fun/interesting. I definitely like the idea since I'm actually doing something similar in what I plan to release eventually as open source that also includes a webUI to help manage hosts, setup clusters, manage & deploy inference "recipes" (there's the similar part). Basically I've been doing approximately the same thing but 1) image level (so what container image it's built on is part of the "recipe") and 2) all shell although I was considering if JSON would be a better interchange format. Was planning to also release that webUI tool as a docker container and then as x86+ARM so that it could run locally on spark or on user's desktop and then control spark via SSH.
> 
> The reason I mention it here when it seems wildly unrelated is because I was thinking of trying to have a git repo that would be an exchange/repository for people's confirmed recipes to make it easier for people to get started. And if I go that route... and profiles are coming this way... it would be good if we more or less standardized representations so that it becomes easy to trend towards a unified ecosystem.
> 
> One thought that would be relevant is if the image could/should be a field in the profile (with some sort of placeholder/default value for local build/image?). Obviously I am leaning that way, but it's a bit more nuanced here. But I like that direction because then if something worked in version XX, you can easily run it with matching version [also why I am going around publishing versioned images...]
> 
> Other thoughts (feel free to reject... I'm just rambling): runtime is currently equivalent of entrypoint in docker; command is then 2nd part which is sort of the first argument; might be easier if command is the executable/first thing and then args is just a text list and we entirely leave interpretation of args to the shell and command. In my WIP stuff, I separated args related to the runtime (VLLM_ARGS) and args related to the model (MODEL_ARGS) but that might be contrived/unnecessary. I would maybe add top level items for tp and ep even if it becomes redundant to the args because I foresee a future where people can browse profiles/recipes via a dashboard, so filtering by parallelism would be good. Even with what I said about command/runtime, I do like listing the runtime (again, thinking about future filtering). Other idea: add env field. configuration of some models/advanced options might require environment variables to be configured.

I'm working on the same thing. It's basically a tool which deploys/runs/stops/removes all know community and NVIDIA playbooks on your spark. It also provides a recipe for you to create/extend the playbooks. I'm close to finish the first cut, I'll share with you for comments.

eugr:
@raphaelamorim - I like the idea, but I wonder if it would be much cleaner from both implementation and architecture standpoint to implement profiles as bash files. This way we just copy them to the head node and execute them there. Lots of upsides:

1. Can introduce some extra logic/steps, like setting environment variables more easily, even download some extra data. 
2. Can avoid any potential escape character / quotes issues.
3. Works with parameters that require `=` sign and not just space (I don't think there are such parameters in vLLM, but still).
4. No extra dependencies.

We can still add some metadata there. 

Alternatively (or additionally), we can think of a profile as a complete replacement for command line arguments and just make it possible to specify args to launch-cluster there, including exec. I'd recommend YAML over JSON though for better readability and lesser chance of mishandling quotes.

raphaelamorim:
> @raphaelamorim - I like the idea, but I wonder if it would be much cleaner from both implementation and architecture standpoint to implement profiles as bash files. This way we just copy them to the head node and execute them there. Lots of upsides:
> 
> 1. Can introduce some extra logic/steps, like setting environment variables more easily, even download some extra data.
> 2. Can avoid any potential escape character / quotes issues.
> 3. Works with parameters that require `=` sign and not just space (I don't think there are such parameters in vLLM, but still).
> 4. No extra dependencies.
> 
> We can still add some metadata there.
> 
> Alternatively (or additionally), we can think of a profile as a complete replacement for command line arguments and just make it possible to specify args to launch-cluster there, including exec. I'd recommend YAML over JSON though for better readability and lesser chance of mishandling quotes.

hmmm makes a lotta sense, particularly the remote execution part. Check again and comment on it, please. Let me know if you agree with the direction

eugr:
@raphaelamorim - better, but still more complicated than it could be. 

I'd propose to rename it to `--launch-script` from `--profile` to explicitly state the purpose and structure bash files like this:

```bash
#!/bin/bash
# PROFILE: OpenAI GPT-OSS 120B
# DESCRIPTION: vLLM serving openai/gpt-oss-120b with FlashInfer MOE optimization

vllm serve openai/gpt-oss-120b \
        --host 0.0.0.0 \
        --port 8888 \
        --enable-auto-tool-choice \
        --tool-call-parser=openai \
        --reasoning-parser=openai_gptoss \
        --gpu-memory-utilization 0.70 \
        --enable-prefix-caching \
        --load-format fastsafetensors
```

This way it's just a simple script that can be directly run on a head node and can include as much logic and be as flexible as possible.

And while using mods to init profiles is tempting, this should follow exec rules, like blocking/non-blocking mode depending on script parameters. 

So, basically, what needs to happen is that if `--launch-script` is specified, we copy it to the container as "exec-script.sh", `chmod +x` it there (similar to how we do it with run.sh for mods) and then set `COMMAND_TO_RUN` to `./exec-script.sh`.

raphaelamorim:
> @raphaelamorim - better, but still more complicated than it could be.
> 
> I'd propose to rename it to `--launch-script` from `--profile` to explicitly state the purpose and structure bash files like this:
> 
> ```shell
> #!/bin/bash
> # PROFILE: OpenAI GPT-OSS 120B
> # DESCRIPTION: vLLM serving openai/gpt-oss-120b with FlashInfer MOE optimization
> 
> vllm serve openai/gpt-oss-120b \
>         --host 0.0.0.0 \
>         --port 8888 \
>         --enable-auto-tool-choice \
>         --tool-call-parser=openai \
>         --reasoning-parser=openai_gptoss \
>         --gpu-memory-utilization 0.70 \
>         --enable-prefix-caching \
>         --load-format fastsafetensors
> ```
> 
> This way it's just a simple script that can be directly run on a head node and can include as much logic and be as flexible as possible.
> 
> And while using mods to init profiles is tempting, this should follow exec rules, like blocking/non-blocking mode depending on script parameters.
> 
> So, basically, what needs to happen is that if `--launch-script` is specified, we copy it to the container as "exec-script.sh", `chmod +x` it there (similar to how we do it with run.sh for mods) and then set `COMMAND_TO_RUN` to `./exec-script.sh`.

please take a look. obvious next step is supporting other runtimes

eugr:
Looks good to me at the first glance. Will review and test later tonight or tomorrow and merge if there are no issues. Thanks for your contribution!

eugr:
I wonder if we are creating more problems with this than we solve? 

There is so much variance in how you might want to run the models - different people would want different gpu utilization limits, maybe different context sizes, definitely different ports and even host (e.g. if want to serve to localhost only). 

And then some models require mods to run, different container builds. Clustered and solo.

I've seen your draft PR about autobuild, but I wonder whether this added complexity actually solves anything better than a good documentation/recipes page? 

I've always been a big proponent of Unix philosophy - a tool should have a relatively narrow scope and not turn into a swiss army knife. I would like to not deviate too much from this :)

Just some late night thoughts... Not saying we shouldn't do it, but I'd like to discuss more.

eugr:
One idea is to keep this `--launch-script` parameter in `./launch-cluster.sh`, but also create a separate script for those who want "one click solution", call it `run-model-recipe.sh` (or even in Python, so more complex logic could be implemented) or something like that. It can have customizable parameters with sane defaults for port/host/cluster/solo, the model profile/recipe can be more complex and include dependencies (mods, cluster mode, container build).

The script would do what you propose in autobuild PR, generate a launch script and call `launch-cluster.sh` with appropriate parameters and a launch script.

raphaelamorim:
> One idea is to keep this `--launch-script` parameter in `./launch-cluster.sh`, but also create a separate script for those who want "one click solution", call it `run-model-recipe.sh` (or even in Python, so more complex logic could be implemented) or something like that. It can have customizable parameters with sane defaults for port/host/cluster/solo, the model profile/recipe can be more complex and include dependencies (mods, cluster mode, container build).
> 
> The script would do what you propose in autobuild PR, generate a launch script and call `launch-cluster.sh` with appropriate parameters and a launch script.

Yeah, I was also thinking about it. The ever growing number of options and complicated parsing on launch-cluster.sh almost cries out to be pushed out to a separate script that add more orchestration of launch-cluster consolidated options and adding a mechanism for users to copy and paste recipes and just execute them without knowing the details of the command line. I can see the number of contributions increasing just on the recipes front.

Added this:

feature: Add recipe-based one-click model deployment system

Introduces a YAML recipe system for simplified model deployment:

- run-recipe.py: Main script handling build, download, and launch
- run-recipe.sh: Bash wrapper for dependency management
- recipes/: Pre-configured recipes for common models
  - glm-4.7-flash-awq.yaml: GLM-4.7-Flash with AWQ quantization
  - glm-4.7-nvfp4.yaml: GLM-4.7 with NVFP4 (cluster-only)
  - minimax-m2-awq.yaml: MiniMax M2 with AWQ
  - openai-gpt-oss-120b.yaml: OpenAI GPT-OSS 120B with MXFP4

Key features:
- Auto-discover cluster nodes with --discover, saves to .env
- Load nodes from .env automatically on subsequent runs
- cluster_only flag for models requiring multi-node setup
- build_args field for Dockerfile selection (--pre-tf, --exp-mxfp4)
- Solo mode auto-strips --distributed-executor-backend ray
- --setup flag for full build + download + run workflow
- --dry-run to preview execution without running

eugr:
Yeah, this looks much better now! A couple of things so far:

1. Let's rename profiles folder to examples or example-launch-scripts as it's not the main way to do recipes now.
2. Changes to launch-cluster.sh can be greatly simplified as we only need a launch script on the head node, not on workers. 

I'll review the Python code next.

eugr:
don't push any more changes, I'm working on it on my end, will merge shortly

eugr:
Merged, thanks for the contribution!

PR #28, opened 2026-02-03
Autobuild Feature

raphaelamorim:
Feature: add DEFAULT_CONTAINER directive and --auto-build flag

Add support for profiles to specify their required Docker image via
the DEFAULT_CONTAINER directive. This allows profiles to declare which
container image they need (e.g., vllm-node-mxfp4 for MXFP4 support).

New features:
- Parse `# DEFAULT_CONTAINER: image-name` from profile scripts
- Add --auto-build flag to automatically build missing images
- Auto-build invokes build-and-copy.sh with --copy-to for worker nodes
- Interactive prompt when image is missing (without --auto-build)

The --auto-build flag requires --launch-script since it reads the
DEFAULT_CONTAINER directive from the profile to determine which image
to build. The -t flag still overrides any profile-specified container.

Updated profiles with DEFAULT_CONTAINER:
- example-vllm-minimax.sh: vllm-node
- vllm-glm-4.7-nvfp4.sh: vllm-node-tf5
- vllm-openai-gpt-oss-120b.sh: vllm-node-mxfp4

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -3,6 +3,8 @@
 # Default Configuration
 IMAGE_NAME="vllm-node"
 DEFAULT_CONTAINER_NAME="vllm_node"
+IMAGE_EXPLICITLY_SET="false"
+AUTO_BUILD="false"
 # Modify these if you want to pass additional docker args or set VLLM_SPARK_EXTRA_DOCKER_ARGS variable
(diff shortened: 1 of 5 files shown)


eugr:
Let's close this one, since all the work is not incorporated into another PR.

PR #39, opened 2026-02-11
Add *_ARCH_LIST Dockerfile args

SurealCereal:
Added `TORCH_CUDA_ARCH_LIST` and `FLASHINFER_CUDA_ARCH_LIST` Dockerfile args to support architectures besides 12.1a

This repo makes builds easier and more reliable for Blackwell 12.0 also.  I left the default `12.1a`.
AFAIK the args need to be re-declared per stage.


eugr:
Thanks, I'll add this with proper support in the build script

eugr:
Implemented in https://github.com/eugr/spark-vllm-docker/commit/3b1e49dcb039f2c28357f29e88a93f4b3c4385cf
Use `--gpu-arch` flag, e.g. ``./build-and-copy.sh --gpu-arch 12.0f`

PR #47, opened 2026-02-15
Add GDN all-mode prefix caching mod for Qwen3-Coder-Next

SilviuSavu:
## Summary

- Adds `mods/gdn-prefix-caching/` mod that applies [vLLM PR #26807](https://github.com/vllm-project/vllm/pull/26807) (GatedDeltaNet all-mode prefix caching) with three additional fixes:
  - `ChunkGatedDeltaRule` CustomOp `forward_native()`/`forward_cuda()` missing `return_intermediate_states` parameter (TypeError crash)
  - `Qwen3NextForCausalLM.get_mamba_chunk_size()` returning 64 (GDN kernel chunk size) instead of the 2048 Mamba2 default
  - `HybridAttentionMambaModelConfig` patched to check model class for chunk_size before falling back to ModelConfig default
- Adds `recipes/qwen3-coder-next-fp8-local.yaml` for single GB10 with prefix caching enabled

## Results on DGX Spark (GB10)

| Metric | align mode (before) | all mode (after) |
|--------|-------------------|-----------------|
| Block size | 1072 | **1088** |
| Cache hit rate | 0.9% | **66.3%** |
| TTFT (warm, ~8.6K prompt) | ~2.7s | **~0.9s (42x speedup)** |
| GPU cache blocks | 1205 | **2273** |

## Test plan

- [x] Server starts with `--enable-prefix-caching --mamba-cache-mode all`
- [x] Block size correctly computed as 1088 (not 2048)
- [x] First inference request succeeds (no TypeError crash)
- [x] Repeated prompts show prefix cache hits in `/metrics`
- [x] Varying user messages with same system prompt show cache reuse


eugr:
I don't know, based on the original PR, there are some accuracy issues with this. 

Also, I'm not a big fan of replacing whole files here as we target vLLM main branch where everything changes rapidly. I'd rather have a diff-based mod that fails harmlessly and doesn't introduce any changes than the one that would break functionality next time they change something in the code (and there are a lot of changes in that part of the codebase now).

I'm getting pretty good prefix cache rates with Claude Code - seeing around 65% in my tests.

SilviuSavu:
yeah, a complete merge makes no sense. I can't get my prefix caching that high. don't know why :(

eugr:
Are you running the latest vLLM build? There have been quite a few commits recently that broke this model in different ways, but it all seems to be resolved now.

PR #55, opened 2026-02-20
Speed up autodiscovey

Clancey:
The Autodiscovery is very slow, I looked at how nvidia does it in their autodiscovery setup https://raw.githubusercontent.com/NVIDIA/dgx-spark-playbooks/refs/heads/main/nvidia/connect-two-sparks/assets/discover-sparks

And applied the same principals. I left the old autodiscovery setup as a fallback.  But now autodiscovery happens in milliseconds compared to seconds.

### DIFF
--- a/autodiscover.sh
+++ b/autodiscover.sh
@@ -112,57 +112,98 @@ detect_nodes() {
     fi
 
     echo "Auto-detecting nodes..."
-    
-    if ! command -v nc &> /dev/null; then
-        echo "Error: nc (netcat) not found. Please install netcat."
-        return 1
-    fi
-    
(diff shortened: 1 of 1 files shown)


eugr:
I used this method before, but it causes all sorts of glitches if you have IP addresses assigned to the second interface. Works OK for ssh setup as it doesn't care through what interface you route things, but will affect the cluster setup.

If you set up your inter-spark network with static IPs from the same class C subnet, the discovery also takes milliseconds.

eugr:
This is the commit if you are interested :) 
https://github.com/eugr/spark-vllm-docker/commit/db5c4439058adb484969177928cbe02a836741b6

PR #56, opened 2026-02-20
Allow Recipies to install as a service

Clancey:
I am not sure if this is something you're interested in, but sometimes I like the recipe and it's easier just to keep running it that way. So, I added a way to install/uninstall it as a service. So I can keep them running on the machine or cluster.

### DIFF
--- a/recipes/README.md
+++ b/recipes/README.md
@@ -191,6 +191,11 @@ Launch options:
   -t, --container IMAGE       Override container from recipe
   --nccl-debug LEVEL          NCCL debug level (VERSION, WARN, INFO, TRACE)
 
+Systemd service (auto-start on boot):
+  --install-service           Install and enable a systemd service for the recipe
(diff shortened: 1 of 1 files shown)


eugr:
I'll review tomorrow

raphaelamorim:
@Clancey, thanks for you contribution. We're moving execution/lifecycle of recipes to https://github.com/scitrera/sparkrun. All the cluster orchestration code will be managed by it. Please take a look and it should be really straightforward to push this feature.

SeanThomasWilliams:
I think the systemd restart logic + docker (non) restart logic will make this systemd service problematic unless the docker container entrypoint is refactored to be a one-shot script which prepares the environment, applies mods and runs the recipe. The main issue is that the docker container itself has the `sleep infinity` entrypoint. 

https://github.com/eugr/spark-vllm-docker/blob/main/launch-cluster.sh#L663-L673

PR #57, opened 2026-02-21
add benchmark in recipe w/ llama-benchy

dscain:
## Summary

Once the LLM is running in Spark, benchmark setup and execution is done manually and can vary between users. This can lead to less precisely reproducible benchmark results.

This PR adds recipe benchmark execution so benchmark setup can be done automatically and benchmark conditions can be recorded directly in the recipe for more reproducible and scientifically precise comparisons.

If a recipe defines a `benchmark` block and sets `benchmark.enabled: true`, benchmark execution is triggered automatically after the model server is up.

## Usage

```bash
./run-recipe.sh ./recipes/openai-gpt-oss-tiny.yaml -d
```

## New recipe block

```yaml
benchmark:
  enabled: true
  framework: llama-benchy
  model: HuggingFaceTB/SmolLM2-135M-Instruct
  args:
    pp: [128]
    depth: [0, 128, 256]
    enable_prefix_caching: true
    save_result: test.md
```

## Behavior

- Benchmark runs only when:
  - `benchmark` block is present, and
  - `benchmark.enabled: true`.
- Currently supported framework: `llama-benchy`.
- Runner waits for `/v1/models`, prints the endpoint payload, resolves the first model id, and runs the benchmark with that resolved model.
- `benchmark.args` are passed generically to CLI flags (`snake_case` -> `--kebab-case`) and forwarded to the benchmark framework.

## Additional recipe example

I also added a more realistic benchmark profile to Nemotron with benchmark disabled by default to allow more discussion and feedback without interfering with existing setups:

```yaml
benchmark:
  enabled: false
  framework: llama-benchy
  args:
    pp: [2048]
    depth: [0, 4096, 8192, 16384, 32768, 65535, 100000]
    enable_prefix_caching: true
    save_result: benchmark_NVIDIA-Nemotron-3-Nano-30B-A3B-NVFP4.md
```


eugr:
Good idea, and timely, since I'm working on automated nightly builds that would run all recipes, automatically benchmark them and check for any regressions.

However, needs some extra work:

1. It seems like it was unintentional and probably caused by a coding agent you use, but there are some code blocks and comments removed, for example, from run-recipe.sh, that should stay there.
2. Running a benchmark should be optional and controlled by a command line parameter to run-recipe.sh.
3. It would be nice to be able to override benchmark parameters.

dscain:
Hello, thank you for the feedback.

1 Removed code blocks: that was actually my mistake (not the coding agent). I accidentally mixed up files while cleaning things up. I've restored the removed code and comments.
2. CLI control for benchmark execution: that is a good point. Initially, benchmark execution was controlled only via `benchmark.enabled` in the recipe.
I’ve now added CLI support so it can also be toggled directly from `run-recipe.sh` with `--benchmark false`.
3. Overriding benchmark parameters: currently, benchmark.args are translated generically into CLI flags for benchy. I intentionally kept parameters inside YAML to avoid coupling run-recipe.sh to a specific benchmark framework and prevent flag collisions between recipe runner parameters and benchmark parameters. That said, I agree that overriding parameters (especially for nightly automation) would be useful.  One possible approach could be something like `--benchmark-override '{pp: 2048}'`.  I'm happy to iterate on the best interface here. If you agree, I suggest we keep this PR focused on the current functionality and discuss/design the override mechanism in a separate follow-up PR to keep the scope clean.

eugr:
OK, agreed on 3, but let's reverse the logic in the CLI flag - by default the benchmark is disabled, you need to specify `--benchmark` flag in order to run it.

dscain:
Currently it is:
1. In Yaml, If no "benchmark" section on recipe.yaml, then "do not run benchmark".
2. If "benchmark" present, but "enabled" not present, assume "false". Do not run benchmark.
3. If "Benchmark" and "enabled" present, only run benchmark if "enable: true"

Therefore, all recipes, as they are now, will not run benchmark because they do not have a "benchmark" block. Nemotron has a benchmark block, but I set it to "enabled: false". The only recipe that has "enable: true" is "smollm2-135m-instruct.yaml", and it is a small model  I find useful for testing.

If you do chose to use "--benchmark", it defaults to "false". To have enable benchmark, you really need "--benchmark true". So the flag alone without "true" or "false" will not enable the benchmark. See this block:

```python
    benchmark_enabled = recipe["benchmark"]["enabled"]
    if args.benchmark is not None:
        benchmark_enabled = args.benchmark == "true"
    # Keep recipe config in sync so downstream benchmark helpers
    # (e.g., run_benchmark.run_recipe_benchmark) honor CLI override.
    recipe["benchmark"]["enabled"] = benchmark_enabled
```

I believe this implements the desired behavior for --benchmark you mentioned above.

dbotwinick:
That's a cool idea. It makes sense to consolidate benchmarking. I wonder about making benchmarking part of the recipe though...  Because for something like spark-arena, provided you don't try to run larger than available context window, etc., you should always use a standardized benchmark. However, it would make sense more for regression testing that a simpler, possibly model-specific benchmark would make sense. Although, it might just make sense to have a standardized benchmark that e.g. tests contexts up to 32768 with tg and varying concurrency for regression tests? So custom benchmarks would basically be for custom use -- I do feel like having benchmarking "recipes/profiles/something" makes sense but I'm actually not sure if they should be part of the recipe itself?  (Just thinking out loud here.)  I've been pondering similar things for sparkrun, so I wanted to engage in discourse.

dscain:
Having benchmark profiles could be helpful for a classes of use cases or types of models indeed. The idea to put in the recipe is to avoid decoupling. Having benchmark.yaml in a separate file would raise the question as how the model was tested (e.g., which parameters), which is a problem we have today, since I cannot be sure I am testing on my Sparks the same way others did in spark-arena. There is also a chance that some benchmarks perform better with certain recipes and/or mods.

Another option is a benchmark that includes a reference to the recipe.yaml, but then again we decouple and start talking about a folder with yamls, complicating the setup. The benchmark in recipe.yaml could help at least give some pointers on the expected use of that vllm in that recipe setup, as a secondary feature.

Anyway, maybe we could merge this one so we at least start capturing benchmark parameter used by others, even if not in its final form. I am using the benchmark-in-the-yaml feature locally often.

dbotwinick:
Interesting point about considering decoupling. I hadn't really seen that as an issue -- but I do see how that could become messy if a lot of things are in flux at the same time.. maybe we do both? Meaning that separate from my other work on the concept -- perhaps each model can (optionally) define a default benchmarking profile for that model as part of the recipe. We can still do benchmarking profiles, etc. -- but each recipe can define a custom, default benchmark for that model.

My two cents:
- Remove enabled from the benchmark block, it should only be determined by CLI. I think CLI flag should be the only thing that can trigger benchmarking / I don't think whether a benchmark is enabled should be inherent to the recipe file.
- model should not be repeated in the benchmark block because it should always be able to be pulled from the outer recipe data -- even if using a custom model served name -- that should be able to be pulled from the recipe data

```
# Optional benchmark configuration
benchmark:
  framework: llama-benchy          # default llama-benchy
  args:
    pp: [2048]                     # default [2048]
    depth: [0]                     # default [0]
    enable_prefix_caching: true    # default true
    save_result: test.md           # default test.md
```

Thoughts?

dscain:
What is in the benchmark section as "model" actually is the model id that will be exposed via the API /v1/models   When performing a request, you need to specify the model id, and that can be different from the other fields specified in the recipe, as it actually happened to me when running benchy the first few times.

About the benchmark enabled only via CLI, I am ok with either way to close the PR. Any thoughts from @eugr  ?

dbotwinick:
Obviously on eugr for what goes here, but I think that we should standardize on convention for setting served model name as part of the recipes -- otherwise the problems with model ID consistency applies not only to benchmarking (as you observed) but to _everything_ that might involve running that model. 

In the `glm-4.7-flash-awq.yaml` recipe, the convention is that `served_model_name` is in the defaults section and then referenced in the command. For the others, it is not mentioned in the defaults or command, which would make the HF model name and served name be the same. So I think that the model ID should pull from defaults.served_model_name if defined otherwise assume its the HF model. 

And in sparkrun, I use defaults.served_model_name as the convention for vllm, sglang, and llama.cpp. It always uses the same pattern exactly to avoid confusion as to how the model will be served.

eugr:
> * Remove enabled from the benchmark block, it should only be determined by CLI. I think CLI flag should be the only thing that can trigger benchmarking / I don't think whether a benchmark is enabled should be inherent to the recipe file.

Yes, that was my point actually. Do not run a benchmark unless there is a CLI flag to do so.

eugr:
@dbotwinick - about served_model_name and HF name.

These are separate concepts. HF name is a "true" model ID - they ID used to pull the model from HF. I think we should always use HF model name as model ID.

`served_model_name` directly translates to the same flag in vLLM and SGLang and defines how the model is exposed in the API by the backend. I don't think we should even try to force a certain model name to recipe users, and serve the model using its HF id by default. 

If users want they can always override based on their needs. E.g., I started to run all coding models as simply "coder" so I don't have to reload Claude Code plugin in VSCode every time I switch to a different model.

llama-benchy uses the same principle. `--model` **always** refers to model ID on HF, so it could download an appropriate tokenizer. `--served-model-name` allows to override to send a different ID to the backend.

eugr:
@raphaelamorim - this PR has evolved into a whole discussion. What do you think about including benchmarking parameters into the recipe?

dbotwinick:
@eugr, that's fair re: HF name vs served_model_name. I'm open to that direction as far as the recipe spec is concerned. It does seem tidy to me to primarily rely on HF model name. Just want to make sure that people can have a mechanism to enforce names for their convenience -- and that we consider how to make that a fluid experience for people -- but it is tidy if we operate off of the HF model ID as the standard.

On topic for here, for adding a default model-specific benchmarking profile into recipe files, my opinion is that the model shouldn't need to be respecified inside of the benchmark block/definition. That was my main point actually. The convention for default served name was actually somewhat incidental to that. 

I actually think we're mostly on the same page. My two points that were feedback on the PR were: 
(1) enabled shouldn't be part of yaml definition and that 
(2) model ID shouldn't be repeated in the yaml benchmark block. 

I was pointing out with defaults.served_model_name that, even if using a non-standard name, we have the necessary plumbing to determine the correct inputs for the benchmarking call. So I think we can discuss/debate on how we want to handle user overrides for served model name -- and defaulting to HF name makes sense -- and we should come to a consensus on that for having a good unified recipe spec. But either way, I don't think we should repeat model under benchmarking if we include benchmarking in the recipe spec.

Also, unrelated, for claude code: I've been using https://github.com/musistudio/claude-code-router on and off. Makes it easy to switch models at a routing/proxy layer between claude code and models.

dscain:
All good, moved benchmark to exclusively "--benchmark true".

eugr:
No need for "true", just the presence of "--benchmark" is enough

dscain:
My intention was to have true/false so it would be a cleaner syntax in build workflows, like "--benchmark $IS_BENCHMARK_ENABLED" and also previewing benchmark parameters overrides. In any case, I have modified to require "--benchmark" (no true/false)

dscain:
After the previous PR merge, this PR had conflicts, which I just fixed. It is possible to merge now? Anything missing in the PR? I believe the "--benchmark" was addressed. Thank you!

eugr:
@dscain - sorry for the delay, will review and merge this today or tomorrow.

PR #62, opened 2026-02-24
support daemon mode for ACTION == exec

dbotwinick:
currently using daemon mode (passed through via recipes) is only respected for clusters; when using a single host or explicit --solo, it stays attached.

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -404,8 +404,8 @@ if [[ "$ACTION" == "status" ]]; then
 fi
 
 # Trap signals
-# Only trap if we are NOT in daemon mode, OR if we are in exec mode (always cleanup after exec)
-if [[ "$DAEMON_MODE" == "false" ]] || [[ "$ACTION" == "exec" ]]; then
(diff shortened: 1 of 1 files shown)


eugr:
Thanks! There was a reason why I launched exec in blocking mode - because `bash -c` (non-interactive) won't load ~/.bashrc where the docker-side cluster script stores all cluster-related environment variables. It doesn't matter in solo mode, but can cause weird bugs in clusters. I wanted to refactor this, but it wasn't a highly requested feature.

I'll see how I can incorporate your changes without causing issues in the cluster.

dbotwinick:
I figured it's a bit of an edge case, but it came up. The fix seemed to work for me, but I didn't go through exhaustive testing on it. Not exactly urgent/critical.

dscain:
I had the same issue and added a fix, although yours look more elaborate https://github.com/eugr/spark-vllm-docker/pull/57

mattcurf:
Appreciated, folks!

eugr:
Sorry guys for the delay, I made some changes that didn't break proper cluster setup. Not very elegant, but working. Maybe will refactor in the future, time permitting.

PR #64, opened 2026-02-24
QWEN 3.5 models : create new mod for update transformers & qwen 3.5 recipe

giraudremi92:


eugr:
Is it really necessary? You can build the container with `--tf5` flag to enable Transformers v5.

eugr:
It works even without `--tf5` as long as you have a fresh build.

giraudremi92:
Thanks for your comment 
I will close my PR

PR #69, opened 2026-02-25
Fix mod for Qwen3.5-35B-A3B-FP8 

giraudremi92:
Fix the Error  TypeError: unsupported operand…type(s) for |: ‘list’ and ‘set’
For Qwen3.5-35B-A3B-FP8 TF5
@eugr

### DIFF
--- a/mods/fix-qwen3_5-moe-tf5/run.sh
+++ b/mods/fix-qwen3_5-moe-tf5/run.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+# Fix incompatibility between vLLM's qwen3_5_moe config and transformers 5:
+# ignore_keys_at_rope_validation is a list but transformers 5 uses | operator (set union)
+set -e
+
+FILE="/usr/local/lib/python3.12/dist-packages/transformers/modeling_rope_utils.py"
+
+echo "Patching $FILE..."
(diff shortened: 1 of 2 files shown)


eugr:
wait, this one runs without any issues on recent vLLM build (non-TF5).

eugr:
I run it with non-tf5 version like this, and no errors:

```bash
./launch-cluster.sh \
 --solo \
exec vllm serve Qwen/Qwen3.5-35B-A3B-FP8 \
--max-model-len 262144 \
--gpu-memory-utilization 0.7 \
--port 8888 --host 0.0.0.0 \
--load-format fastsafetensors \
--enable-prefix-caching \
--enable-auto-tool-choice \
--tool-call-parser qwen3_coder
```

You need a recent vLLM build though.

giraudremi92:
I ll try it without tf5

eugr:
@giraudremi92 - did it work?

stavinsky:
> @giraudremi92 - did it work?

even it it works all other models works fine with tf5. I've made similar patch for myself so I can I have only one container for everything
But instead of sed magic i've used more common diff patch. 
may be we should have something like this existing temporary for a while?

giraudremi92:
It works on TF4 new build  ! 
@stavinsky yes it depends of the build we use.

image

image

@eugr should i close the PR ?

eugr:
I'll close the PR then.

PR #70, opened 2026-02-26
Fix rsync failure if destination dir doesn't exist

hoesing:
Add `--mkpath` arg to model copy rsync command in order to create the destination `.cache/huggingface/hub` folder if it doesn't already exist.


eugr:
Looks good, merging

PR #76, opened 2026-03-04
Fix shell quoting for exec command arguments

laudney:
## Summary
- Fix improper quoting of exec command arguments in `launch-cluster.sh` and `run-recipe.py`
- Arguments containing special characters (e.g. JSON strings) were passed unquoted, causing breakage for commands like:
  ```
  --speculative-config '{"method":"qwen3_next_mtp","num_speculative_tokens":2}'
  ```
- Uses `printf %q` in `launch-cluster.sh` and `shlex.quote()` in `run-recipe.py` to properly escape arguments

## Tested
- Verified exec command with JSON argument containing special characters works correctly

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -115,7 +115,7 @@ while [[ "$#" -gt 0 ]]; do
             fi
             ACTION="exec"
             shift
-            COMMAND_TO_RUN="$@"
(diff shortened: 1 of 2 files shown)


eugr:
thank you! I will test and merge today!

PR #84, opened 2026-03-06
Prevent 169.254.x.x fallback when setting fix IP address

oparoz:
To force the use of the static IP address we've chosen to be assigned to the interface, it's safer to disable the fallback to avoid problems down the line


oparoz:
Should be squashed before merging if I don't get to it ;)

eugr:
Thanks, merged!

Issue #86, opened 2026-03-07
Fails to download dependencies during build


SeraphimSerapis:
I suggest some logs that could help debug what is exactly failing (e.g., what is curl trying to download at that moment).

Any proposed steps to recover/fix this?

Logs below:

```
./build-and-copy.sh
Downloading flashinfer_cubin-0.6.5-py3-none-any.whl...
######################################################################## 100.0%
Downloading flashinfer_jit_cache-0.6.5-cp39-abi3-manylinux_2_28_aarch64.whl...
######################################################################## 100.0%
Downloading flashinfer_python-0.6.5-py3-none-any.whl...
######################################################################## 100.0%
Downloading ...
curl: (3) URL rejected: Malformed input to a URL function
Failed to download  — removing other downloaded files.
Restoring previous flashinfer wheels...
mv: cannot stat './wheels/.backup-download-flashinfer/.flashinfer*': No such file or directory
Download failed — using existing local FlashInfer wheels.
Downloading vllm-0.17.0rc1.dev114+g2a00d3241.d20260306.cu131-cp312-cp312-linux_aarch64.whl...
######################################################################## 100.0%
Downloading ...
curl: (3) URL rejected: Malformed input to a URL function
Failed to download  — removing other downloaded files.
Restoring previous vllm wheels...
mv: cannot stat './wheels/.backup-download-vllm/.vllm*': No such file or directory
Download failed — using existing local vLLM wheels.
Building runner image with command: docker build -t vllm-node --build-arg BUILD_JOBS=16 --build-arg TORCH_CUDA_ARCH_LIST=12.1a --build-arg FLASHINFER_CUDA_ARCH_LIST=12.1a .
[+] Building 2.3s (15/15) FINISHED                                                                                                                                                                                           docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                                   0.0s
 => => transferring dockerfile: 11.68kB                                                                                                                                                                                                0.0s
 => resolve image config for docker-image://docker.io/docker/dockerfile:1.6                                                                                                                                                            0.7s
 => CACHED docker-image://docker.io/docker/dockerfile:1.6@sha256:ac85f380a63b13dfcefa89046420e1781752bab202122f8f50032edf31be0021                                                                                                      0.0s
 => [internal] load metadata for nvcr.io/nvidia/pytorch:26.01-py3                                                                                                                                                                      1.4s
 => [internal] load .dockerignore                                                                                                                                                                                                      0.0s
[...]

PR #87, opened 2026-03-07
fix: skip empty lines in wheel download read loop

SeraphimSerapis:
This PR addresses #86 

This line seems to cause trouble:
https://github.com/eugr/spark-vllm-docker/blob/9dc09bd04b6420a3b3c155c05bf2542dbb0dbdd7/build-and-copy.sh#L179

Proposed fix: Add a guard to skip empty lines (e.g. trailing newlines) in the while-read loop to prevent try_download_wheels from breaking on unexpected blank input.


eugr:
Thanks!

PR #93, opened 2026-03-11
super nemotron mod & recipe for NVIDIA-Nemotron-3-Super-120B-12B-NVFP4

giraudremi92:


hashmesan:
Recipe works, thank you! via Asus Ascent

eugr:
@giraudremi92 @hashmesan - I've adjusted the recipe, removed unnecessary parameters and added prefix caching, so pls update from main.

Sggin1:
While testing the Marlin path from this PR, also did a llama.cpp
sm_121 native build as a baseline comparison. Sharing results in
case useful context for the PR.

Nemotron-3-Super 120B on single Spark:
- llama.cpp sm_121 native (Q4_K, 66GB): 14.43 t/s
- Ollama sm_120 (Q4_K_M, 86GB): 14.18 t/s
- Generation is memory-bandwidth bound at this scale — sm_121
  target doesn't move the needle until NVFP4 native dispatch works

The Marlin path in this PR (~16-17 t/s) is still the best available
option for the NVFP4 checkpoint. CUDA 13.2 + new CUTLASS sm_121
support looks like the real unlock — watching closely.

Build recipe and full benchmarks:https://gist.github.com/Sggin1/cd21ed471c861e814a85925ee04dfed6

eugr:
Thanks for sharing!

PR #97, opened 2026-03-14
Add --server mode for vLLM as container PID 1

RobTand:
## Summary

- Adds `--server` flag to `launch-cluster.sh` and `run-recipe.py` that runs vLLM as the container's main process (PID 1) instead of `sleep infinity` + `docker exec`
- Docker's `--restart unless-stopped` policy is applied by default, so if vLLM crashes the container automatically restarts the model
- Also adds `--restart <policy>` flag for explicit control of Docker restart policy independent of server mode

## Motivation

In the current solo + daemon mode, if vLLM crashes the container stays up but the model goes offline until manually restarted. This requires external watchdog scripts to detect and recover from crashes. Server mode makes Docker handle this natively.

## Usage

```bash
# Via run-recipe.sh
./run-recipe.sh recipes/qwen3.5-122b-int4-autoround.yaml --solo --server

# Via launch-cluster.sh directly
./launch-cluster.sh --solo --server --launch-script my-script.sh

# With custom restart policy
./run-recipe.sh recipes/my-model.yaml --solo --server --restart on-failure:5
```

## How it works

1. Container starts with a wait loop (same pattern as mod application)
2. Mods are applied and launch script is copied into the container
3. The wait loop `exec`s the vLLM command, replacing PID 1 with vLLM
4. On crash, Docker restarts the container; mods and script persist in the container's writable layer, so vLLM starts immediately on restart

## Test plan

- [ ] Verify `--server` mode starts vLLM as PID 1 (`docker exec vllm_node ps -p 1`)
- [ ] Verify crash recovery: `docker exec vllm_node kill 1` should trigger automatic restart
- [ ] Verify `launch-cluster.sh --solo stop` cleanly stops and removes the container
- [ ] Verify existing daemon mode (`-d`) is unaffected
- [ ] Verify `--restart` flag works independently of `--server`

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -375,13 +375,18 @@ cleanup() {
     # Stop Head
     echo "Stopping head node ($HEAD_IP)..."
     docker stop "$CONTAINER_NAME" >/dev/null 2>&1 || true
-    # Remove container if it exists (needed when not using --rm, e.g. server mode)
-    docker rm "$CONTAINER_NAME" >/dev/null 2>&1 || true
+    # Remove container if not using --rm (e.g. server mode with restart policy)
+    if [[ -n "$RESTART_POLICY" ]]; then
+        docker rm "$CONTAINER_NAME" >/dev/null 2>&1 || true
+    fi
 
     # Stop Workers
     for worker in "${PEER_NODES[@]}"; do
         echo "Stopping worker node ($worker)..."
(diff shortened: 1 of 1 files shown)


RobTand:
@eugr big fan of your work on this project. I wanted to add this feature instead of relying on a cron watchdog script. This is my first contribute to open source

RobTand:
I see the open PR re sparkrun. I'm gonig to close this and use sparkrun instead.

PR #102, opened 2026-03-16
feat(recipes): Intel/MiniMax-M2.5-int4-AutoRound with tokenizer mod

kilork:
# Fix MiniMax-M2.5 tokenizer drift and compatibility issues with Transformers v5

## Problem

The default tokenizer_class (e.g., MiniMaxTokenizer or GPT2Tokenizer) in tokenizer_config.json causes significant tokenization errors when used with transformers v5.0+. Specifically, it leads to "tokenization drift" (skipped characters, incorrect spacing, and corrupted byte sequences) because the high-level Python wrappers apply incompatible normalization rules to the MiniMax architecture.

## Solution

To resolve this, the mod overrides the tokenizer class to use the raw Rust-based engine directly. Also the PR adds new recipe for Intel AutoRound quantization of MiniMax-M2.5 (Intel/MiniMax-M2.5-int4-AutoRound) which is using this mod.

## Changes:

1. Override Tokenizer Class: Replaced tokenizer_class with TokenizersBackend in tokenizer_config.json. This forces the engine to bypass transformers Python logic and use the raw tokenizer.json rules directly via the Rust tokenizers library.
2. Explicit Chat Template: Since TokenizersBackend does not provide a default chat format, the patch requires an external Jinja template (compatible with MiniMax/ChatML) to be passed via the --chat-template flag in vLLM. This is taken directly from model.

## Possible enhancements

- I have not done deep performance tests, as I am more interested in correct output than speed.
- If we can know model name during mod deployment time, mod can be tuned to support any MiniMax-M2.5 model affected (including NVFP4), as they share same config settings apparently.
- Not clear, if --chat-template here is the right one, in my tests with OpenCode I have not seen any issues in normal flow, but dcp (opencode plugin to manage context size) tags were not recognized.

### DIFF
--- a/recipes/minimax-m2.5-int4.yaml
+++ b/recipes/minimax-m2.5-int4.yaml
@@ -20,7 +20,6 @@ cluster_only: true
 # No mods required
 mods:
   - mods/fix-qwen3.5-autoround
-  - mods/fix-qwen3.5-chat-template
   - mods/minimax-m2.5
 
 # Default settings (can be overridden via CLI)


kilork:
Hmm, I guess it is not need, closing with note for myself.

PR #103, opened 2026-03-16
Add docker --network arg to common build flags

apairmont:
This PR is a non-breaking change.

Certain corporate DNS server make you jump through hoops when using docker bridged.  Simplest solution is to pass `--network host` to the docker build command.  This fixes timeouts on DNS lookup for apt and pip.

I added it via COMMON_BUILD_FLAGS so it's easier to maintain.

### DIFF
--- a/build-and-copy.sh
+++ b/build-and-copy.sh
@@ -21,6 +21,7 @@ PRE_TRANSFORMERS=false
 FULL_LOG=false
 BUILD_JOBS="16"
 GPU_ARCH_LIST="12.1a"
+NETWORK_ARG=""
 WHEELS_REPO="eugr/spark-vllm-docker"
 FLASHINFER_RELEASE_TAG="prebuilt-flashinfer-current"
 VLLM_RELEASE_TAG="prebuilt-vllm-current"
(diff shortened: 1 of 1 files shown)


eugr:
That's a good one, thanks! Will merge.

oparoz:
Thank you so much for this. I had such restrictions in place and hat to "hack" the build file :)

eugr:
Thanks for flagging that, I should probably add a dedicated section in README describing this issue. How did the issue manifest?

oparoz:
In my case, I created a default rule on all Docker subnets to not be able to connect to the Internet. That way only the containers I place on a specific network can go play outside.
I never realised this build script needed access from a Docker subnet until I implemented the change and then Identified where the problem was in the build script

apairmont:
I am on a corporate network that seems to prevent DNS access from within containers (perhaps related to DHCP guarding but I'm not certain). DNS servers simply refuse to respond, but network access still worked by IP address, so apt and pip couldn't resolve hostnames and would timeout. Simplest solution was to use host network during build.

PR #105, opened 2026-03-18
Fix: Correctly insert multi-node args after model name in --no-ray mode

kitch2400:
## Problem
When using `--no-ray` mode with `--launch-script`, the multi-node arguments were being inserted between `vllm serve` and the model name, causing vLLM to fail with "model should be provided as the first positional argument" error.

## Solution
Fixed the `make_node_script()` function to insert arguments **after** the model name instead of before it.

### Changes
- Fixed sed command in `make_node_script()` to properly insert args after model name
- Added support for model paths with spaces (quoted paths)
- Handles both multi-line commands and single-line commands
- Added clarifying comments to `exec_no_ray_cluster()`

## Testing
Tested successfully with multi-line commands, model paths with spaces, and single-line commands.

## Impact
Enables the `--no-ray` flag to work correctly with `--launch-script` for multi-node vLLM clusters without Ray overhead.

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -559,7 +559,10 @@ make_node_script() {
 
     local tmp; tmp=$(mktemp /tmp/vllm_node_script_XXXXXX.sh)
     grep -v -- '--distributed-executor-backend' "$script_path" > "$tmp"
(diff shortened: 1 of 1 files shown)


eugr:
Pushed my own changes that fix this behavior.

PR #110, opened 2026-03-18
Remove run-cluster-node.sh script copy and permission commands from Dockerfile.mxfp4

voloszad:
--mxfp4 image build was failing as the run-cluster-node.sh was removed.


eugr:
oops, thanks! Although MXFP4 builds are still a bit broken, I believe.

PR #124, opened 2026-03-22
Validate cached wheels before skipping download

kreuzhofer:
## Summary

- Fixes a bug where corrupt wheel files (e.g. containing a GitHub 404 "Not Found" response) would never be re-downloaded because the commit-hash or timestamp checks passed without verifying file integrity
- Adds zip magic-byte validation (`PK\x03\x04` header check) at both early-exit paths in `try_download_wheels`
- When corrupt wheels are detected, removes them and their commit marker so the normal download path runs

## Root Cause

If a previous download received a 404 from GitHub, the response body ("Not Found", 9 bytes) was saved as the `.whl` file. On subsequent runs:
1. The commit-hash check matched (`.flashinfer-commit` was still valid) → skip download
2. The timestamp check passed (corrupt files were newer) → skip download
3. The runner `docker build` failed with `unable to locate the end of central directory record`

## Test plan

- [x] Simulated corrupt wheels (`echo "Not Found" > *.whl`), verified detection and automatic re-download
- [x] Verified valid wheels pass the check (tested against real vllm wheel)
- [x] Full `docker build` succeeds end-to-end after fix
- [x] `./tests/test_recipes.sh -v` passes (46/46, 1 pre-existing unrelated failure)


eugr:
Thanks, this failure mode should be very rare, but still need to guard against it.
I'll review and merge tomorrow if no issues.

PR #131, opened 2026-03-24
Add podman support via CONTAINER_RT environment variable

sjug:
Add support for podman as a drop-in alternative to docker for all container operations. All scripts now use a `CONTAINER_RT` environment variable (defaults to `docker`) instead of hardcoded `docker` commands. This applies to `launch-cluster.sh`, `build-and-copy.sh`, and `run-recipe.py`.

If a container was stopped but not removed (e.g. crash), docker run would fail because the container name is taken. There was no cleanup of these stale/failed containers before starting. Replaced `--rm` for container run command with explicit `rm` before start — `--rm -d` removes the container before logs are retrievable. Added `rm` to the cleanup function for consistent teardown on both head and worker nodes.

### DIFF
--- a/build-and-copy.sh
+++ b/build-and-copy.sh
@@ -4,6 +4,9 @@ set -e
 # Start total time tracking
 START_TIME=$(date +%s)
 
+# Container runtime (docker or podman)
+CONTAINER_RT="${CONTAINER_RT:-docker}"
+
 # Default values
 IMAGE_TAG="vllm-node"
 IMAGE_TAG_SET=false
@@ -101,7 +104,7 @@ add_copy_hosts() {
 get_remote_image_id() {
     local host="$1"
     local image="$2"
-    ssh "${SSH_USER}@${host}" "docker image inspect --format '{{.Id}}' ${image}" 2>/dev/null
+    ssh "${SSH_USER}@${host}" "$CONTAINER_RT image inspect --format '{{.Id}}' ${image}" 2>/dev/null
 }
 
 copy_to_host() {
@@ -109,7 +112,7 @@ copy_to_host() {
     echo "Loading image into ${SSH_USER}@${host}..."
     local host_copy_start host_copy_end host_copy_time
     host_copy_start=$(date +%s)
-    if cat "$TMP_IMAGE" | ssh "${SSH_USER}@${host}" "docker load"; then
+    if cat "$TMP_IMAGE" | ssh "${SSH_USER}@${host}" "$CONTAINER_RT load"; then
         host_copy_end=$(date +%s)
         host_copy_time=$((host_copy_end - host_copy_start))
         printf "Copy to %s completed in %02d:%02d:%02d\n" "$host" $((host_copy_time/3600)) $((host_copy_time%3600/60)) $((host_copy_time%60))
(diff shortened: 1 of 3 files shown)


eugr:
Thanks! Does the build work properly for both source and quick (wheels) builds with Podman?

I remember I had to add `--format docker` to `podman build` to make it work with multistage builds on Fedora before. Don't know if it's still needed.

eugr:
Also, launch parameters could be different, and I remember mount points had to be defined a bit differently for correct permissions mapping.

sjug:
I've [put together a new PPA](https://launchpad.net/~sejug/+archive/ubuntu/podman/+packages) for the current `podman` version & deps as Ubuntu Noble doesn't carry a modern version. That's what I've been using to test functionality with your scripts and I've not had any issues or command changes. 

Multi-stage builds work without `--format docker`; that flag only changes the image manifest format.

If you have a RHEL based distro with selinux then you'll need the `:z`/`:Z` flag for the volume mounts.

sjug:
@eugr Are you looking for some code changes? Or what outcome do you want?

eugr:
@sjug - sorry, I'm finalizing mesh networking support, so I'm way behind on PRs at the moment. I'll get to yours, I promise. 

Does it work with Podman from stock OS packages?

sjug:
Yes and no, the stock `podman` version is very old so it needed a few extra changes.

1. `--gpus` all did not work with old podman
    - The launch script uses Docker-style:
      ```bash
        --gpus all
      ```
    - We had to use CDI syntax:
      ```bash
        --device nvidia.com/gpu=all
      ```
2. Generated NVIDIA CDI spec was too new
    - nvidia-cdi-refresh.service generated:
      ```bash
        /run/cdi/nvidia.yaml
      ```
    - But it used:
      ```yaml
        cdiVersion: 0.7.0
        additionalGids:
      ```
    - Podman 4.9.3 rejected both:
        - unknown field "additionalGids"
        - invalid version "0.7.0"
3. Had to downgrade the CDI spec for old podman
    - Removed `additionalGids`
    - Changed:
      ```yaml
        cdiVersion: 0.7.0
      ```
      to:
      ```yaml
        cdiVersion: 0.5.0
      ```
4. Had to launch non-privileged
    - With CDI + rootless podman, privileged mode hit:
      ```text
        /dev/dri/card1: permission denied
      ```
    - So the working launch needed:
      ```bash
        --non-privileged
      ```

Final working command looked like:

```bash
  VLLM_SPARK_EXTRA_DOCKER_ARGS="--device nvidia.com/gpu=all" \
  CONTAINER_RT=podman \
  ./run-recipe.sh openai-gpt-oss-120b --solo --non-privileged -d
```

peshovec:
Hi, @sjug 
Probably not directly related to the PR, however can be related.
I use podman for inference (still use the docker build, and then export).
When an os upgrade switch the new version of cdi - I had to upgrade podman by uisng your repo.
There were some issues, including apparmor ones (the apparmor mostly with pasta), but it went well. And most important, the next full OS upgrade went smoothly. 
Saying that, because in my opinion using recent podman version is the way to go. E.g. probably there will be need for prerequisite to switch the podman ppa to yours. This way the cdi and etc. can stay unmodified

PR #134, opened 2026-03-25
Fix: stabilize Intel/Qwen3.5-397B-A17B-int4-AutoRound loading via fastsafetensors deduplication

stergiom:
## Summary
Fix for `Intel/Qwen3.5-397B-A17B-int4-AutoRound` checkpoint compatibility with the `fastsafetensors` loader.

## Problem
The official `Intel/Qwen3.5-397B-A17B-int4-AutoRound` checkpoint has duplicate tensor keys leaked between `shard_39` and `shard_40`. While the default vLLM loader silently overwrites these duplicates, the `fastsafetensors` loader strictly enforces unique keys and crashes with: `key ...qweight must be unique among files`

Additionally, on GB10 unified memory machines (128GB shared CPU/GPU RAM), the default NCCL broadcast path in `fastsafetensors` causes hard OOM panics during loading. This is caused by massive temporary network buffers being allocated when Rank 0 attempts to scatter the 5GB chunks across the cluster.

## Solution
Two-part fix:

### 1. Deduplication Utility (`dedup-safetensors.py`)
- Scans sharded safetensors checkpoints for duplicate tensor keys.
- Enforces "last shard wins" semantics (matches standard PyTorch/HF sequential loader behavior) to preserve the most recent quantization pass.
- Updates `model.safetensors.index.json` to reflect the correct, sanitized shard mappings.
- Creates automatic `.bak` backups before modification and verifies integrity post-rewrite.
- Supports `--dry-run` mode for standalone inspection.

**Usage:**
```bash
python3 dedup-safetensors.py /path/to/models/Intel/Qwen3.5-397B-A17B-int4-AutoRound
```

### 2. FastSafeTensors No-Broadcast Mod (`mods/fix-fastsafetensors-no-broadcast/`)
Forces `fastsafetensors` to drop to a `SingleGroup` (independent per-rank loading) instead of using the TP Process Group for NCCL scatter broadcasting.
- **Before:** Rank 0 reads all tensors from a shard (~5GB) and allocates massive NCCL flight buffers in unified memory to broadcast half away → OOM kernel panic.
- **After:** Each rank independently reads the full checkpoint from disk (local NVMe or networked) and quietly keeps its own partition.

### 3. Recipe Update (`recipes/qwen3.5-397b-int4-autoround.yaml`)
- Added `mods/fix-fastsafetensors-no-broadcast` to the automatic mod chain.
- Injected `--load-format fastsafetensors` flag.

## Testing
- Tested on dual-node DGX Spark cluster (GB10 SM121, TP=2).
- Successfully loads full 397B parameters natively with `fastsafetensors` in ~3m.
- 0 OOM crashes during weight loading.

```bash
docker exec -it vllm-head vllm bench serve \
  --backend vllm \
  --model Intel/Qwen3.5-397B-A17B-int4-AutoRound \
  --dataset-name random \
  --random-input-len 512 \
  --random-output-len 128 \
  --num-prompts 500
```

```text
============ Serving Benchmark Result ============
Successful requests:                     500
Failed requests:                         0
Benchmark duration (s):                  1584.69
Total input tokens:                      256000
Total generated tokens:                  64000
Request throughput (req/s):              0.32
Output token throughput (tok/s):         40.39
Peak output token throughput (tok/s):    54.00
Peak concurrent requests:                500.00
Total token throughput (tok/s):          201.93
---------------Time to First Token----------------
Mean TTFT (ms):                          792260.33
Median TTFT (ms):                        792448.99
P99 TTFT (ms):                           1566111.27
-----Time per Output Token (excl. 1st token)------
Mean TPOT (ms):                          46.02
Median TPOT (ms):                        46.10
P99 TPOT (ms):                           47.43
---------------Inter-token Latency----------------
Mean ITL (ms):                           46.02
Median ITL (ms):                         42.98
P99 ITL (ms):                            55.54
==================================================
```

## Files Changed
- `dedup-safetensors.py` (new) - Standalone deduplication utility
- `mods/fix-fastsafetensors-no-broadcast/` (new) - Mod to disable network scattering
- `recipes/qwen3.5-397b-int4-autoround.yaml` - Added loader flags and documentation

## Notes
- The deduplication script is purely standalone and generic; it can be pointed at any corrupted HuggingFace safetensor cache.
- The `no-broadcast` mod is specifically engineered to stabilize GB10 unified memory architectures, but provides stable performance ceilings for any memory-constrained edge deployment.

### DIFF
--- a/dedup-safetensors.py
+++ b/dedup-safetensors.py
@@ -1,0 +1,186 @@
+#!/usr/bin/env python3
+"""
+dedup-safetensors.py — Remove duplicate tensor keys from sharded safetensors checkpoints.
+
+Some quantized checkpoints produce shards with overlapping tensor keys.
+fastsafetensors requires globally unique keys and will error with
+"key ...qweight must be unique among files".
+
+IMPORTANT: This script uses "last shard wins" semantics, NOT "index wins".
+Standard PyTorch/HF loaders process shards sequentially (00001 → 00041).
+If the same key appears in both shard 39 and shard 40, shard 40's version
+silently overwrites shard 39's. This is significant because quantization
+boundary bugs (e.g., Intel AutoRound on the 397B) can leave aborted/garbage
+tensors in an earlier shard while the finalized version lands in a later shard.
+The index.json may point to the EARLIER shard (the garbage), so trusting the
+index would delete the correct data. Instead, we keep the last-writer-wins
+version and update the index to match.
+
+Usage:
+    python3 dedup-safetensors.py /models/Qwen3.5-397B-A17B-int4-AutoRound
+    python3 dedup-safetensors.py /models/... --dry-run
+
+Requires: pip install safetensors torch
+"""
+
+import argparse
+import json
+import shutil
+import sys
+from collections import defaultdict
+from pathlib import Path
+
+from safetensors.torch import load_file, save_file
+
+
+def scan_shards(model_dir: Path):
+    """Scan all shards and find which keys exist in which shards."""
+    index_path = model_dir / "model.safetensors.index.json"
+    if not index_path.exists():
+        print(f"ERROR: {index_path} not found.")
+        sys.exit(1)
+
+    with open(index_path) as f:
+        index = json.load(f)
+
+    weight_map = index["weight_map"]
+
+    # Find all shard files (from index + any extras on disk)
+    shard_files = sorted(set(weight_map.values()))
+
+    # Build map: key -> list of shards that contain it (in shard order)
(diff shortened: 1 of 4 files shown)


eugr:
I wonder if you could look into patching fastsafetensors loader to behave similar to regular safetensor one?
It would be much better solution that we can even embed into the base image, or at least into a mod, vs. having a separate script that we need to run to patch the model first. I guess the script could also be added to the mod, but patching fastsafetensors would be much better.

stergiom:
Yeh, that is a better way of going about it. Will close this for now and take a look between things.

PR #155, opened 2026-04-02
Add custom volume mount support to recipe launches

czardoz:
Allow run-recipe.py to accept repeated --mount-volume flags and pass them through to launch-cluster.sh.

Teach launch-cluster.sh to translate each mount into a docker -v argument and add integration coverage for both the recipe dry-run path and launcher check-config output.

This is useful for e.g when we need to pass `--ssl-key-file` and `--ssl-cert-file` to vLLM.


eugr:
This is useful, thanks. Will review tomorrow.

czardoz:
@eugr can you check this change out if you get a chance? I'm carrying it in my local repo, but I'd love to just have it upstreamed. Happy to make changes etc if you have any feedback.

czardoz:
@eugr Friendly ping, just in case :) Sorry for bugging!

PR #165, opened 2026-04-07
Add Gemma 4 31B FP8 recipe with reasoning + tool parser fixes

kreuzhofer:
## Summary

- New recipe `gemma-4-31b-fp8.yaml` for Google Gemma 4 31B dense model with runtime dynamic FP8 quantization
- New mod `fix-gemma4-reasoning` fixing vLLM #38855 (reasoning parser fails when special tokens are stripped)
- Recipe uses `vllm-node-tf5` container and both `fix-gemma4-reasoning` + `fix-gemma4-tool-parser` mods

## Recipe Details

- **Model:** `google/gemma-4-31B-it` (BF16 base, runtime FP8 quantization via `--quantization fp8`)
- **Container:** `vllm-node-tf5` (transformers>=5 baked in)
- **FP8 KV cache** for additional memory savings
- **Single DGX Spark** (TP=1), 65K context, 0.85 GPU memory utilization
- **Tool calling:** `--tool-call-parser gemma4 --reasoning-parser gemma4`

## Mods Applied

1. **fix-gemma4-reasoning** (new): Fixes #38855 — `extract_reasoning()` receives text with `skip_special_tokens=True` already applied, so `<|channel>` delimiters are missing. Fix re-decodes with `skip_special_tokens=False` when reasoning boundary token IDs are present. Tries PR #38858 diff first, falls back to inline Python patch.

2. **fix-gemma4-tool-parser** (existing): PR #38909 — streaming tool call HTML duplication fix.

## Why runtime FP8 instead of pre-quantized checkpoint?

The `protoLabsAI/gemma-4-31B-it-FP8` checkpoint has a `weight_scale` KeyError in `gemma4.py load_weights` (same class of bug as vllm #38912). Using the base BF16 model with `--quantization fp8` for dynamic online quantization avoids this entirely.

## Test Results

Tested on DGX Spark (GB10):

| Test | Streaming | Result |
|------|-----------|--------|
| Basic chat | No | PASS |
| Basic streaming | Yes | PASS |
| Single tool call | No | PASS — valid JSON |
| Single tool call | Yes | PASS — valid JSON streamed |
| Multi-tool call (2 tools) | No | PASS |
| Agentic multi-step | No | PASS |
| Agentic multi-step | Yes | PASS |
| HTML in streaming tool args | Yes | PASS — no `<<` duplication |
| Tool selection (3 available) | No | PASS — correct tool chosen |
| Full agentic loop (2 tools → results → answer) | Yes | PASS |


dbotwinick:

ChristianWeyer:
Servus Daniel :-D @kreuzhofer

Issue #167, opened 2026-04-07
vllm images can't be built anymore


alankila:
Hi,

I'm a new owner of one of those GB10 machines (ThinkStation PGX). I was kicking the tires and got mediocre performance out of llama.cpp and decided to try getting vllm running. I tried to use the qwen3.5-122b-a10b model at int4, which has a recipe.

The docker build of vllm-node-tf5 image seemed to go fine, but something in the result doesn't appear to be binary compatible. My guess is that it's because it depends on things like "apt-get install" found files from server that can vary as the apt sources vary, and checks out github by branch rather than commit id, so these can change as well, and so the docker build is fundamentally unstable and can't be expected to repeat indefinitely the same.

The patch hunk also fails, which suggests that some maintenance of these recipes might be needed, as it seems like the patch may have become obsolete.

```
Solo mode enabled. Skipping node detection.
Head Node: 127.0.0.1
Worker Nodes: 
Container Name: vllm_node
Image Name: vllm-node-tf5
Action: exec
Starting Head Node on 127.0.0.1...
640e4c720cfaeb1c0ecb68f224c59ef0f5bb6abc0f409b77b0934ce5b4342860
Applying modifications to cluster nodes...
Applying mod 'fix-qwen3.5-autoround' to 127.0.0.1...
  Copying directory content to container...
Successfully copied 4.1kB to vllm_node:/workspace/mods/fix-qwen3.5-autoround/
  Running patch script on 127.0.0.1...
patching file transformers/modeling_rope_utils.py
Hunk #1 FAILED at 648.
1 out of 1 hunk FAILED -- saving rejects to file transformers/modeling_rope_utils.py.rej
Patch is not applicable, skipping...
Applying mod 'fix-qwen3.5-chat-template' to 127.0.0.1...
  Copying directory content to container...
Successfully copied 11.3kB to vllm_node:/workspace/mods/fix-qwen3.5-chat-template/
  Running patch script on 127.0.0.1...
=======> to apply chat template, use --chat-template unsloth.jinja
Copying launch script to head node...
Successfully copied 2.56kB to vllm_node:/workspace/exec-script.sh
Executing command: /workspace/exec-script.sh
Traceback (most recent call last):
  File "/usr/local/bin/vllm", line 4, in <module>
    from vllm.entrypoints.cli.main import main
  File "/usr/local/lib/python3.12/dist-packages/vllm/entrypoints/cli/__init__.py", line 3, in <module>
[...]

PR #177, opened 2026-04-10
fix: make SM121 FP8 patch resilient to upstream changes

SeraphimSerapis:
## Problem

The SM121 (DGX Spark) FP8 fix ([vllm-project/vllm#35568](https://github.com/vllm-project/vllm/pull/35568)) is applied during build via `curl | git apply`. This breaks whenever upstream commits shift the surrounding context lines — even though the actual target code hasn't changed.

The PR was never merged into vLLM and the `git apply` approach is inherently fragile against a fast-moving codebase like vLLM.

## Fix

Replace the inline `curl | git apply` block with a standalone `patches/sm121-fp8-fix.sh` script that uses `sed` for exact string substitution. This makes the patch:

- **Context-independent** — matches exact strings, not surrounding lines
- **Resilient to upstream churn** — survives line shifts from adjacent PRs
- **Self-healing** — silently no-ops if vLLM eventually merges the fix
- **Cleaner** — keeps the Dockerfile readable (3 lines vs 9)

## What the patch does

Applies the same 8 fixes from PR #35568 across 6 files, changing SM120-only checks to SM12x family checks:

| Pattern | Before | After |
|---|---|---|
| Arch check | `arch in [89, 120]` | `arch == 89 or arch // 10 == 12` |
| Capability check | `== 120` | `== 12` (major) |
| CUTLASS guard | `enable_sm120_only` | `enable_sm120_family` |
| Python check | `is_device_capability(120)` | `is_device_capability_family(120)` |

## Breaking commits identified

Two upstream PRs shifted context around the patched code, causing `git apply` to fail:
- `db5d0719` — [#34664](https://github.com/vllm-project/vllm/pull/34664) "Add MXFP8 to Marlin GEMM/MoE" (Apr 1)
- `ab1a6a43` — [#37221](https://github.com/vllm-project/vllm/pull/37221) "Migrate cutlass/scaled_mm_entry.cu torch stable ABI" (Mar 30)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -204,15 +204,9 @@ RUN if [ -n "$VLLM_PRS" ]; then \
         done; \
     fi
 
-# TEMPORARY PATCH for broken FP8 kernels - https://github.com/vllm-project/vllm/pull/35568
-RUN curl -fsL https://patch-diff.githubusercontent.com/raw/vllm-project/vllm/pull/35568.diff -o pr35568.diff \
-    && if git apply --reverse --check pr35568.diff 2>/dev/null; then \
-         echo "PR 35568 already applied, skipping."; \
-       else \
-         echo "Applying PR 35568..."; \
-         git apply -v pr35568.diff; \
-       fi \
-    && rm pr35568.diff
+# SM121 (DGX Spark) FP8 fixes — equivalent to PR #35568 but context-independent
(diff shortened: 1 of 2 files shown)


eugr:
As I explained in #167, this is by design, so closing this for now.

PR #179, opened 2026-04-11
feat: ignore wheel lock and whl files in git

Zambonilli:
Seeing whl and commit files tracked in git when running `./build-and-copy.sh`


eugr:
I actually had a separate gitignore in wheels but it included itself - fixed it now.

PR #190, opened 2026-04-14
Add InstantTensor to runtime dependencies

SeraphimSerapis:
Adds instanttensor as available load format (addresses https://github.com/eugr/spark-vllm-docker/issues/188)

Makes `--load-format instanttensor` available.


eugr:
Nice find! Will merge soon!

PR #199, opened 2026-04-19
Add a recipe option to use local vetted code rather than trust-remote-code.

savagesmc:
Using this repo to run models on a pair of sparks that I use for work. For security reasons, I would like to not run models with the -trust-remote-code option and make sure I have inspected the code that is being 'trusted' before I run it in my work environment. To this end, I added an option into the recipe that lets you do that. Might be useful for some others trying to use sparks in their work environment.


eugr:
Wouldn't you achieve the same by just using HF_HUB_OFFLINE=1 and pre-downloaded model? It should just use code from the model directory in the cache without trying to fetch anything remotely....

PR #207, opened 2026-04-22
Add GB10 MoE config install mod

RPHllc:
## Summary

- add `mods/install-gb10-moe-config`
- install a tuned fused MoE kernel config for `Qwen/Qwen3.6-35B-A3B-FP8` on `NVIDIA GB10`
- document the mod in the changelog and in the README mods list

## Why

On GB10, vLLM can fall back to the default Triton MoE config for this shape and print:

`Using default MoE config. Performance might be sub-optimal!`

This mod installs a benchmarked config for:

- `E=256`
- `N=512`
- `dtype=fp8_w8a8`
- `block_shape=[128,128]`

The config was generated on 2026-04-21 via `benchmarks/kernels/benchmark_moe.py` for `Qwen/Qwen3.6-35B-A3B-FP8` on GB10 hardware.

In local testing on GB10, this avoided the fallback and improved throughput by more than 20%.

## Usage

```bash
./launch-cluster.sh --apply-mod mods/install-gb10-moe-config ...
```

Or in a recipe YAML:

```yaml
mods:
  - mods/install-gb10-moe-config
```

## Notes

This PR keeps the mod narrow and explicit for one verified config rather than trying to generalize it prematurely.

## Validation

- `bash -n mods/install-gb10-moe-config/run.sh`
- verified mod asset is installed from the mod directory into vLLM's fused MoE configs path
- updated README changelog and mods documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,21 @@ For periodic maintenance, I recommend using a filter: `docker builder prune --fi
 
 ## CHANGELOG
 
+### 2026-04-22
+
+#### GB10-tuned MoE config for Qwen3.6-35B-A3B-FP8
+
+Added a new mod `mods/install-gb10-moe-config` that installs a tuned fused MoE kernel config for `Qwen/Qwen3.6-35B-A3B-FP8` on `NVIDIA GB10` systems.
+
+This avoids the default Triton fallback (`Using default MoE config. Performance might be sub-optimal!`) for the tuned shape (`E=256`, `N=512`, `dtype=fp8_w8a8`, `block_shape=[128,128]`) and improved throughput by about 30% during local testing on GB10 hardware.
+
+To use it, add the mod when launching:
+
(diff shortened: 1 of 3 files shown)


SeraphimSerapis:
This is very good work. I'm trying to do the same but for TP=2. Sadly random crashes in the later batches seem to make that more difficult than I was hoping for. Any recommendations?

RPHllc:
> This is very good work. I'm trying to do the same but for TP=2. Sadly random crashes in the later batches seem to make that more difficult than I was hoping for. Any recommendations?

I suggest changing N=512 to N=256

I can't verify or optimize it further, I have only one DGX-spark

SeraphimSerapis:
@RPHllc I addressed this by writing this simple helper: https://github.com/SeraphimSerapis/vllm-tune

Cheers!

PR #213, opened 2026-04-26
fix: update patches for current vLLM and transformers versions

torrotitans:
## Summary

Fixes two patch failures that occur when building the `vllm-node-tf5` container image.

### 1. `mods/fix-qwen3.5-autoround/transformers.patch`
- **Problem:** Patch targeted line 648, but `transformers` was refactored (code moved to line 722)
- **Fix:** Updated line number and context (added `self.` prefix)

### 2. `mods/gpu-mem-util-gb/gpu_mem.patch`
- **Problem:** Hunk #2 at line 239 failed (`LLM.__init__` signature shifted)
- **Fix:** Updated line numbers and context

## Verification
Both patches verified with `patch --dry-run` against current source.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 .env
-build-metadata.yaml.roo
+build-metadata.yaml
+.roo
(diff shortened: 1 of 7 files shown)


eugr:
FYI, transformers patch is not needed anymore, the underlying issue has been fixed in vLLM main.

PR #223, opened 2026-05-01
YAML syntax check: 'container' key was duplicated.

ttsiodras:
Was testing recipes/nemotron-3-super-nvfp4.yaml - and upon opening it in my vim, the yaml language server spotted a small typo: the key 'container' was duplicated. Might as well fix it :-)


ttsiodras:
Side note: I got 15 tg/s in single node GX10 (--solo), 24 tg/s in 2x GX10 cluster.

PR #236, opened 2026-05-08
Add recipe for MiniMax-M2.7-NVFP4

laudney:
## Summary
- Add vLLM serving recipe for MiniMax-M2.7 using the
  nvidia/MiniMax-M2.7-NVFP4 quantization
- Uses Ray distributed backend with 2-way tensor parallelism and
  the CUTLASS MoE backend tuned for NVFP4
- fp8 KV cache, prefix caching, and FlashInfer TRT-LLM allreduce
  backend (`VLLM_FLASHINFER_ALLREDUCE_BACKEND=trtllm`) for
  efficient TP allreduce on NVFP4 GEMMs
- Reuses existing minimax_m2 tool-call and reasoning parsers

## Test plan
- [x] Model serving verified on spark3:8080 with chat completions
  endpoint
- [x] Tool-call and reasoning parser output confirmed working

### DIFF
--- a/recipes/minimax-m2.7-nvfp4.yaml
+++ b/recipes/minimax-m2.7-nvfp4.yaml
@@ -1,0 +1,51 @@
+# Recipe: MiniMax-M2.7-NVFP4
+# MiniMax M2.7 model with NVFP4 quantization (CUTLASS MoE backend)
+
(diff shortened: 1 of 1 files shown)


dbotwinick:

PR #253, opened 2026-05-19
Yarn extended context

csabakecskemeti:
1M and 500k YARN scaled qwen3.6-35B recipes
(tested on a dual DGX Spark setup)


csabakecskemeti:
doing more testing

PR #258, opened 2026-05-21
Install audio decode deps (soundfile, librosa, av) in runner image

s4l4x:
### What
Adds `soundfile`, `librosa`, and `av` to the runner stage so the image can decode audio input.

### Why
vLLM is built from source here without the `[audio]` extra, so neither `soundfile` nor `av` (PyAV) end up in the runner image. As a result every audio request fails in vLLM's `load_audio` with `400 Invalid or unsupported audio file` — it tries `soundfile` first, falls back to `av`, and raises when both are missing. Image and video inputs are unaffected (video uses the OpenCV loader backend, which is already present), so the gap is audio-specific and easy to miss.

This matters for omni models like Nemotron-3-Nano-Omni, which advertise audio understanding but can't receive it without a decoder in the image.

### Change
One dedicated runner-stage layer:
- `soundfile` (libsndfile) — vLLM's primary audio decode path
- `av` (PyAV/ffmpeg) — fallback decoder; also backs `resample_audio_pyav`
- `librosa` — rounds out vLLM's `audio` extra (mel features)

Placed after the existing extra-deps layer so the flashinfer/vLLM wheel builds stay cache-hot — it's a runner-only rebuild. All three have `aarch64` binary wheels.

### Testing
Built and ran on a DGX Spark (GB10, SM_121, CUDA 13.0, aarch64). Before: audio requests 400'd in `load_audio`. After: Nemotron-3-Nano-Omni transcribes a WAV audio fixture correctly. Image/video unchanged.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -343,6 +343,17 @@ ENV PATH=$VLLM_BASE_DIR:$PATH
 RUN --mount=type=cache,id=uv-cache,target=/root/.cache/uv \
     uv pip install ray[default] fastsafetensors instanttensor
 
+# Audio modality support (iris senses).
+# vLLM is built from source above without the [audio] extra, so soundfile AND av
+# are both absent — every audio request 400s in load_audio ("Invalid or
+# unsupported audio file"), while image/video work (video uses the OpenCV loader
+# backend). soundfile (libsndfile) is vLLM's primary audio decode path; av
+# (PyAV/ffmpeg) is the fallback and also backs resample_audio_pyav; librosa
+# rounds out vLLM's `audio` extra. aarch64 binary wheels exist for all three.
+# Ref: iris plans/eval-harness.md v1.6.1 (senses modality × provider matrix).
+RUN --mount=type=cache,id=uv-cache,target=/root/.cache/uv \
+    uv pip install soundfile librosa av
+
 # Fix NCCL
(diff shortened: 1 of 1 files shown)


volschin:
NVIDIA also loads decord2 for av handling in omni. Perhaps worth to include also?

isolitude:
I'm considering modifying build-and-copy.sh to support specifying custom Dockerfiles. As @volschin mentioned, different models often require distinct packages or environments. For instance, decord2 might be needed for omni models, conch-triton-kernels for qwen-3.5-397b, vllm-omni for glm-image, and deepseek-v4 requires a third-party vLLM repository alongside a specific NCCL version.

Merging all of these into our current single Dockerfile isn't sustainable. It would be much cleaner to separate them into different Docker images based on the target model's requirements.

s4l4x:
Understood, and makes sense though I'm not even sure it's worth it. The biggest issue I had was just knowing that audio wasn't working by default. A note on the README would have saved me some time.  

If you did want to go with per model variants, a discoverability matrix would be nice:

| Variant | Capability | Verified Models | Diff |
|---------|------------|-----------------|------|
| [Link](#) | Audio | Nemotron 3 Nano Omni, ... | [Link](#) |

Feel free to close and if you'd like I can propose a note regarding audio

PR #263, opened 2026-05-28
fix: pin torch via --override during wheel install to prevent CPU torch downgrade

Dogacel:
I've faced this issue while trying to run some image using `sparkrun` on a DGX Spark.

```shell
sparkrun run @spark-arena/e8e202f4-c7af-4aaa-87c4-ef2ad9cfba59
```

I got

```shell
File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.12/dist-packages/vllm/platforms/cuda.py", line 21, in <module>
      import vllm._C  # noqa
      ^^^^^^^^^^^^^^
  ImportError: libtorch_cuda.so: cannot open shared object file: No such file or directory".
```

Looking closely at the runtime deps:

```shell
    cuda:      13.2
    python:    3.12.3
    torch:     2.10.0+cpu
    vllm:      0.21.1rc1.dev292+g97e4022c6.d20260526
```

The torch is somehow downgraded to the CPU verison.

We've traced this to flashinfer-python's unpinned `Requires-Dist: torch` causes uv to re-resolve torch when installing the locally-built wheels, downgrading the cu130 build to PyPI's CPU wheel. The resulting image fails at runtime with:

This patch captures the already-installed torch version via `python -c` and passes it as a `--override` to uv, keeping the CUDA build through both the wheel install and the final extras install.

AI usage disclosure: Changes are described and reviewed by me & applied by a coding agent.


bornmw:
the image is broken without this, please merge

vedcsolution:
I hit the same failure on DGX Spark after the latest vLLM wheel update (`torch 2.10.0+cpu`, missing `libtorch_cuda.so`). Your root-cause analysis matches what I see: local wheel/final dependency installs re-resolve `torch` away from the CUDA build.

One smaller upstream-friendly variant to consider: use uv native torch backend selection globally in the Docker stages instead of generating override files:

```dockerfile
ENV UV_TORCH_BACKEND=cu130
```

Then keep the existing explicit initial torch install with `--index-url https://download.pytorch.org/whl/cu130` if desired. The env var protects all later `uv pip install` calls in the stage, including:

- `uv pip install /workspace/wheels/*.whl`
- the `--tf5` wheel install branch
- final extras like `ray[default] fastsafetensors instanttensor`

This avoids future regressions if another dependency starts declaring/transitively resolving `torch`.

I would also add a final runner assertion so the image build fails fast if CPU torch slips in again:

```dockerfile
RUN python3 - <<'PY'
import os
import torch

libtorch_cuda = os.path.join(os.path.dirname(torch.__file__), "lib", "libtorch_cuda.so")
assert torch.version.cuda is not None, f"Expected CUDA torch, got {torch.__version__}"
assert os.path.exists(libtorch_cuda), f"Missing {libtorch_cuda}"
print(f"torch {torch.__version__} cuda {torch.version.cuda}")
PY
```

`Dockerfile.mxfp4` has the same pattern (`uv pip install /mount/wheels/*.whl` and final deps), so it likely needs the same guard.

If you prefer keeping this PR as override-based, it still fixes the immediate issue; the main advantage of `UV_TORCH_BACKEND=cu130` is that it covers all later/future `uv` resolutions without duplicating override setup.

vedcsolution:
Update after testing this locally on DGX Spark:

I tested the `ENV UV_TORCH_BACKEND=cu130` idea. It fixes the CPU-wheel part, but it is not enough by itself: during local wheel install uv still downgraded the pinned runtime from:

```text
torch 2.11.0+cu130
```

to:

```text
torch 2.10.0+cu130
```

So your override approach is needed to preserve the exact torch build that vLLM was built against.

The local patch that worked for both `vllm-node` and `vllm-node-tf5` was:

```dockerfile
echo "torch==$(python3 -c 'import torch; print(torch.__version__)')" > /tmp/torch-override.txt && \
if [ "$PRE_TRANSFORMERS" = "1" ]; then \
    echo "transformers>=5.0.0" > /tmp/tf-override.txt && \
    cat /tmp/torch-override.txt /tmp/tf-override.txt > /tmp/wheel-override.txt && \
    uv pip install /workspace/wheels/*.whl --override /tmp/wheel-override.txt; \
else \
    uv pip install /workspace/wheels/*.whl --override /tmp/torch-override.txt; \
fi
```

Plus a final assertion:

```dockerfile
RUN python3 - <<'PY'
import os
import torch

libtorch_cuda = os.path.join(os.path.dirname(torch.__file__), "lib", "libtorch_cuda.so")
assert torch.__version__ == "2.11.0+cu130", f"Expected torch 2.11.0+cu130, got {torch.__version__}"
assert torch.version.cuda is not None, f"Expected CUDA torch, got {torch.__version__}"
assert os.path.exists(libtorch_cuda), f"Missing {libtorch_cuda}"
print(f"torch {torch.__version__} cuda {torch.version.cuda}")
PY
```

Validation on node `192.168.8.121` after rebuilding/copying both images:

```text
vllm-node:latest      2.11.0+cu130 13.0  libtorch_cuda.so=True
vllm-node-tf5:latest  2.11.0+cu130 13.0  libtorch_cuda.so=True
```

Image IDs from my local rebuild:

```text
vllm-node:latest      sha256:87417b31184073a44634a827154daea02e1cfe98d5636fc80889d18a7f994a1c
vllm-node-tf5:latest  sha256:5edb6c59430f1081a6580e5d70f2f0928a3e41aaea3d8f7fd926bb90b8f9813d
```

So correction to my previous comment: `UV_TORCH_BACKEND=cu130` is useful as extra protection against CPU torch, but it does not preserve the required torch version. The torch override is the critical fix.

lucas-nicolas:
applied the fix locally after getting :

```
File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/python3.12/dist-packages/vllm/platforms/cuda.py", line 21, in <module>
      import vllm._C  # noqa
      ^^^^^^^^^^^^^^
  ImportError: libtorch_cuda.so: cannot open shared object file: No such file or directory".
```

the proposed fix worked on my end as well

eugr:
Thank you!

johnmschoonover:
> Thank you!

And thank you for the quick merge! Excited to get back on main tomorrow!!

PR #269, opened 2026-05-31
Drop cu130 index and switch to PyPi PyTorch

xor2k:
Since PyTorch version 2.11, `--index-url https://download.pytorch.org/whl/cu130` is not necessary anymore to get cu130 builds, as the standard PyTorch has `cu130` included. However, those packages from PyPi don't have `+cu130` endings, so a `uv pip list` shows `torch==2.11.0` instead of `torch==2.11.0+cu130`. However, `torch.__version__` still yields `2.11.0+cu130`, so builds will fail. Therefore, I "relaxed" the version checking a little bit (which it completely fine for PyTorch >= 2.11 because it includes cu130 anyway).

<s>There is however a catch: I did not yet remove `--index-url https://download.pytorch.org/whl/cu130`, because there seems to be a difference with the PyPi version of `torch` and the one provided by that index: around 51 TPS for `qwen/Qwen3.6-35B-A3B-FP8` instad of around 54-55 TPS. Also, upgrading from PyTorch 2.11 to PyTorch 2.12 seems to degrade performance to around 52 TPS, both for the PyPi-version as well as for the version on the index.</s>

For PyTorch 2.12 I've rebuild `flashinfer` and `vllm` from scratch, since otherwise the `build-and-copy.sh` will fail.

I'll benchmark the PyPi PyTorch version against the one on the PyTorch cu130 index a little bit further and maybe file a bug report. But other than that, this PR will insure that future PyTorch versions can simply be enabled by changing a version number.

Alternatively, this can be considered such a minor performance regression that PyTorch 2.12 from PyPi still gets enabled, then I can modify my PR accordingly.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -53,7 +53,7 @@ RUN apt update && \
 
 # Additional deps
 RUN --mount=type=cache,id=uv-cache,target=/root/.cache/uv \
-     uv pip install torch==2.11.0 torchvision torchaudio triton --index-url https://download.pytorch.org/whl/cu130 && \
+     uv pip install torch==2.11.0 torchvision torchaudio triton && \
      uv pip install nvidia-nvshmem-cu13 "apache-tvm-ffi<0.2" filelock pynvml requests tqdm
 
 # Configure Ccache for CUDA/C++
@@ -794,7 +794,7 @@ ARG PRE_TRANSFORMERS=0
 
 # Install deps
 RUN --mount=type=cache,id=uv-cache,target=/root/.cache/uv \
-     uv pip install torch==2.11.0 torchvision torchaudio triton --index-url https://download.pytorch.org/whl/cu130 && \
+     uv pip install torch==2.11.0 torchvision torchaudio triton && \
      uv pip install nvidia-nvshmem-cu13 "apache-tvm-ffi<0.2"
 
 # Install wheels from host ./wheels/ (bind-mounted from build context — no layer bloat)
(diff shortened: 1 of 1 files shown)


xor2k:
Just checked: the PyTorch packages are identical in PyPi and the PyTorch cu130 index. I've also checked more packages, will continue my search these days, got a bunch of ideas. But I have some ideas already how to narrow down the potential speed difference.

xor2k:
Good news: no performance degradiation. What definitely helps with measurement is to reboot the machine, then 55 TPS+ for Qwen3.6-35B-A3B-FP8 are achieved just as before.

xor2k:
Some more info on what I did:
```
docker image rm vllm-node:latest
docker system prune -a
```
to really make sure the caches are gone and then build with
```
./build-and-copy.sh --rebuild-flashinfer --rebuild-vllm --vllm-ref a4ac746
```
The `--vllm-ref` is pretty random, but fixed and one that does not crash the build with an error and for testing. `--rebuild-flashinfer --rebuild-vllm` might not be necessary anymore once this PR gets pulled as new versions will be rebuild with PyTorch 2.12 by default.
```
curl -sN -X POST "http://localhost:11434/v1/chat/completions" \
  -H "Authorization: Bearer YOUR-API-KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "qwen/Qwen3.6-35B-A3B-FP8",
    "messages": [
      {
        "role": "user",
        "content": "Generate some lengthy text"
      }
    ],
    "stream": true
  }' | sed -u 's/^data: //' \
| jq -rj '
    select(. != "[DONE]") |
    if (.choices | length) == 0 then
      empty
    else
      .choices[0].delta as $d |
      (
        if $d.reasoning_content then
          "\u001b[2m" + $d.reasoning_content + "\u001b[0m"
        elif $d.reasoning then
          "\u001b[2m" + $d.reasoning + "\u001b[0m"
        elif $d.content then
          $d.content
        else
          empty
        end
      )
    end
  '
```

xor2k:
Just rebased

@eugr Anything that speaks against merging this pull request? Sorry for asking directly, but haven't heared anything so far and it seems relatively obvious to me. If upgrading PyTorch to 2.12 is a problem, one can also stick wich PyTorch 2.11 and remove the cu130 index, which would improve dependency tracking significantly. This is also the default installation method since PyTorch 2.11.

As mentioned, I did some benchmarking and when there were differences, I even compared the individual dependency packages (cu130 index and PyPi) and they are identical. But the cu130 index is very limited regarding package selection and pip is only partially capable of handling different indices properly (if that is even possible at all).

xor2k:
Actually, just figured out there is an issue with version pinning outside of the cu130 index that e.g. affects DeepSeekV4 Flash. Update is coming...

xor2k:
Fixed the version pinninig, checked, builds all fine.

xor2k:
Okay, so here the latest state of this PR:
1. PyTorch stays 2.11
2. dropped `--index-url https://download.pytorch.org/whl/cu130` from `uv pip install`
3. Reorganized pinning code to remove duplicates
4. No performance regressions
5. Primary motivation for this PR: use wheels outside of container, don't have `--index-url` interfere with other package installations

@eugr No further known issues from my side, would be more than happy if you could have a look.

PR #278, opened 2026-06-06
Make nccl wheel

xor2k:
Currently, `nccl` is created as a debian package. The main disadvantage is that since pip packages have precedence with dynamic library loading, the pip packages need to be patched with the version from the deb every time they get installed. Therefore, it would be more practical to have a nccl wheel with the patched nccl in the first place. Once that wheel is available, the docker can be fully used for building DGX Spark specific wheels that then can be used outside of the container. <s>I've also added the `override.txt` for easy reuse outside of docker to the `wheels` directory.</s>


xor2k:
<s>For that PyTorch version upgrade for 2.12 also see https://github.com/eugr/spark-vllm-docker/pull/269</s> Removed for better testability, PyTorch 2.12 requires packages to be rebuilt.

xor2k:
<s>Just noticed there is still a little bug: the wrong nccl version is used, I'll fix that tomorrow</s> Done.

xor2k:
So, ultimately, it needed a dedicated `nccl` build phase in `build-and-copy.sh`. Otherwise, the main parts still happen in the Dockerfile. I've also validated with `uv pip freeze | grep @` via `./launch-cluster.sh --solo exec bash` that the right (custom) packages have been installed, including `nvidia-nccl-cu13`.

xor2k:
Sorry for closing and reopening, happened by accident.

xor2k:
Just rebased and tested, still works fine.

@eugr what do you think? I've been building wheels (including nccl) and used them outside of the container since quite a while now and it works really well, both on a single machine as well as my 2x DGX Spark cluster.

PR #281, opened 2026-06-07
Add recipe for deepseek-ai/DeepSeek-V4-Flash on 2x DGX Spark (GB10)

FyrbyAdditive:
Adds `deepseek-v4-flash-jasl-gb10`, a recipe for serving [`deepseek-ai/DeepSeek-V4-Flash`](https://huggingface.co/deepseek-ai/DeepSeek-V4-Flash) on a two-node GB10 / sm_121 DGX Spark cluster, plus a CHANGELOG entry under `2026-06-07` with image-build instructions.

## Why a separate recipe

DeepSeek-V4-Flash will not run on consumer spark clones with the stock `vllm-node` image:

1. The model config sets `swiglu_limit=10.0`. vLLM's NVFP4 oracle requires the selected backend to honor that clamp; the only backend in `NVFP4_BACKENDS_WITH_CLAMP` is `FLASHINFER_TRTLLM`, and the trtllm-gen NVFP4 MoE kernel rejects sm_120/121 (it requires Blackwell datacenter family 100).
2. The DeepSeek-V4 sparse-attention indexer hardcodes DeepGEMM calls (`fp8_fp4_mqa_logits` / `fp8_fp4_paged_mqa_logits`). DeepGEMM has no sm_120/121 kernels (issue [deepseek-ai/DeepGEMM#324](https://github.com/deepseek-ai/DeepGEMM/issues/324) is still open).

The recipe therefore depends on jasl's [`codex/ds4-sm120-min-enable`](https://github.com/jasl/vllm/tree/codex/ds4-sm120-min-enable) branch tracked by [vllm-project/vllm#41834](https://github.com/vllm-project/vllm/pull/41834), which:

- adds Triton replacements for the sparse-MLA / FP8 indexer paths (DeepGEMM-free on sm_12x);
- routes MoE through MegaMoE, which externalises the SwiGLU clamp via a runtime `activation_clamp=` parameter and so sidesteps the NVFP4 oracle's clamp-set rejection;
- pins NCCL `v2.30u1` built with `-gencode=arch=compute_121` and symlinks the system libnccl over the python wheel to fix the DGX Spark NCCL load-order hang.

A bespoke image (`ds4-vllm-gb10:pr`) is built from [`jasl/vllm-ds4-sm120-harness/docker/gb10-pr`](https://github.com/jasl/vllm-ds4-sm120-harness/tree/main/docker/gb10-pr); the recipe `container:` field points at that tag. The README CHANGELOG entry walks through the build.

## Notable flag choices

[...]

PR #219, opened 2026-04-29
Add DeepSeek V4 Flash recipe

arthur-drozdov:
## Summary

Adds a DeepSeek V4 Flash recipe for the official `deepseek-ai/DeepSeek-V4-Flash` checkpoint on dual DGX Spark / GB10.

- Builds `vllm-node-dsv4` from `jasl/vllm` at vLLM PR #41834 tag `sm120-pr-41834-stable-preview-20260624` (vLLM `20260625.dev0+g28fef2c70`), with FlashInfer rebuilt and DeepGEMM `nv_dev` pulled in for the MoE path.
- Keeps the vLLM checkout compatible with repositories, branches, tags, SHAs, `refs/heads/...`, and `refs/pull/.../head`.
- Pins NCCL `2.30.7` in both the system and Python runtime paths, then repins CUDA Torch after dependency install.
- Enables the guarded FlashInfer SM120/SM121 **prefill + decode** MLA path (`VLLM_DEEPSEEK_V4_FLASHINFER_SM120_PREFILL/DECODE=1`).
- TP=2, FP8 KV cache, prefix caching, FlashInfer autotune, CUDA graphs (`FULL_AND_PIECEWISE`), safetensors loading.
- Defaults: `max_model_len: 200000`, `gpu_memory_utilization: 0.85`, `max_num_seqs: 2`, `max_num_batched_tokens: 4096`.
- DeepSeek V4 tokenizer, tool-call, and reasoning parsers, with explicit `<think>` / `</think>` markers.
- MTP/speculative decoding, expert parallelism, eager mode, and extra timeout settings left disabled.

## DeepGEMM on GB10 (important)

DeepGEMM's FP8 **linear / attention-o_proj** path does not run on GB10/SM121 today: the `deepseek_v4_fp8_einsum` o_proj asserts at init on the UE8M0 weight scales —

```
RuntimeError: Assertion error (csrc/.../layout.hpp:71): sfb_dtype == torch::kFloat or sfb_dtype == torch::kInt
RuntimeError: Assertion error (csrc/.../layout.hpp:97): sf.size(-2) == ceil_div(mn, gran_mn)
```

The recipe therefore forces:

- `VLLM_USE_DEEP_GEMM=0` — disable DeepGEMM for the FP8 linear / o_proj einsum (falls back to `TritonFp8BlockScaledMMKernel`),
- `VLLM_USE_DEEP_GEMM_E8M0=0` — disable the E8M0 DeepGEMM variant,
[...]

PR #292, opened 2026-06-16
Add DeepSeek-V4-Flash recipe (aidendle94 B12X image) + /workspace fix for third-party images

vedcsolution:
## What

- **launch-cluster.sh**: create `/workspace` in the container before copying the launch script, so images that don't ship a `/workspace` dir (third-party / community vLLM builds) also work. `mkdir -p` is idempotent for the repo's own `vllm-node` image.
- **recipes/deepseek-v4-flash-fp8-aiden.yaml**: new recipe to serve `deepseek-ai/DeepSeek-V4-Flash` (FP8 dense + MXFP4 MoE) on 2x DGX Spark via the community `aidendle94/sparkrun-vllm-ds4-gb10` image (native B12X MXFP4 MoE backend, up to ~1M ctx). Supports Ray and no-Ray.

## Why

The aiden image is a popular community build for DeepSeek-V4-Flash on GB10 (B12X MXFP4 kernels). It ships a custom entrypoint (`dsv4-vllm-entrypoint`) and uses `WORKDIR=/tmp` (no `/workspace`). Two small things were needed to drive it from `launch-cluster.sh`:

1. The launcher copies the per-node script to `/workspace/exec-script.sh`, which failed on this image with *"Could not find the file /workspace in container"*. Fixed by creating the dir first (no-op for images that already have it).
2. Because the launcher runs the command via `docker exec` (which bypasses the image entrypoint) and clears the entrypoint on the idle `sleep infinity` container, the recipe invokes the entrypoint binary explicitly inside the command. It runs with **no** `--keep-entrypoint` (keeping it would make the idle container run `vllm sleep infinity` and crash).

## Tested

2x DGX Spark GB10, TP=2, no-Ray: container starts on both nodes, `/workspace` is created on the third-party image, weights load, `/v1/models` + a chat completion are verified, and the `B12X` MXFP4 MoE backend is selected.

### DIFF
--- a/launch-cluster.sh
+++ b/launch-cluster.sh
@@ -798,6 +798,9 @@ make_node_script() {
 copy_script_to_container() {
     local container="$1"; local script_path="$2"; local label="${3:-node}"
     echo "Copying launch script to $label..."
+    # Ensure /workspace exists: the repo's own image has it, but third-party vLLM images
+    # (e.g. custom community builds) may not. mkdir -p is a no-op when it already exists.
+    docker exec "$container" mkdir -p /workspace || { echo "Error: could not create /workspace in $label"; exit 1; }
     docker cp "$script_path" "$container:/workspace/exec-script.sh" || { echo "Error: docker cp to $label failed"; exit 1; }
     docker exec "$container" chmod +x /workspace/exec-script.sh
 }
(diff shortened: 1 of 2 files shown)


vedcsolution:
Heads-up for triage: there are two other open DeepSeek-V4-Flash recipe PRs — #219 (builds DSv4 into the repo's own image) and #281 (jasl fork, `deepseek-v4-flash-jasl-gb10`). This PR is intentionally a different approach, not a duplicate:

- It uses the community **`aidendle94/sparkrun-vllm-ds4-gb10`** image, whose native **B12X MXFP4 MoE backend** is the fast path on GB10. #281's own notes mention that `VLLM_USE_B12X_MOE` looked "specific to `aidendle94/...`" and that recipe fell back to jasl — this PR is the working b12x path with that image (`Using 'B12X' Mxfp4 MoE backend` is selected at load).
- It also includes a small, generally-useful **`launch-cluster.sh`** change: create `/workspace` before copying the launch script, so third-party images that don't ship a `/workspace` dir (this one uses `WORKDIR=/tmp`) work with the launcher. Neither #219 nor #281 touches the launcher, so this is independent of which recipe you prefer.

Happy to split the `/workspace` launcher fix into its own PR if you'd rather review/merge it independently of the recipe.

SpaceshipCreative:
Nice work!

eugr:
The workspace fix has been implemented independently, so you can remove it from the PR. I'm not sure if this repository should contain recipes that depend on pre-built images though.

Also, as a heads-up, vLLM is planning to merge PR 43477 as an official solution to both GLM-5.1 and DSV4F, so I'm currently testing it, and likely will include it in my builds as it will be much easier to maintain.

PR #319, opened 2026-07-15
Add DeepSeek-V4-Flash-DSpark recipe + SM120 topk fix

bilikaz:
DSpark speculative decoding for DeepSeek-V4-Flash on a DGX Spark cluster. Minimal delta off deepseek-v4-flash.yaml: -DSpark checkpoint + method:dspark.

Includes mods/fix-dspark-sm120-topk: the DSpark draft issues a sparse-MLA decode with topk=256, which FlashInfer's SM120 decode_dsv4 kernel has no compiled bucket for ({128,512,1024}) -> it falls through to the paged kernel and crashes at warmup with 'num_tokens > 64'. The mod rounds the draft's topk to the nearest supported bucket (256->128). Correctness-safe: the draft is a speculator, so the target verifies every proposed token; this only affects accept-length, never output. Verified: clean output, mean accept-length ~4 and ~53 tok/s on code workloads (workload-dependent, per the upstream DSpark notes).

### DIFF
--- a/mods/fix-dspark-sm120-topk/run.sh
+++ b/mods/fix-dspark-sm120-topk/run.sh
@@ -1,0 +1,53 @@
+#!/bin/bash
+# fix-dspark-sm120-topk — let DSpark serve on SM120 (DGX Spark / GB10, sm_121 → sm120).
+#
+# Problem: DeepSeek-V4-Flash-DSpark's draft runs a sparse-MLA decode with topk=256, but
+# FlashInfer's SM120 `decode_dsv4` kernel only has compiled buckets {128, 512, 1024}
+# (see _DECODE_DSV4_DISPATCH in flashinfer/mla/_sparse_mla_sm120.py). topk=256 sits in
+# the gap → the dispatcher falls through to the paged kernel, which asserts
+# `num_tokens > 64` and crashes during sparse-MLA warmup:
+#   tvm.error.InternalError: Check failed: num_tokens > 64 (5 vs. 64) :
+#   Decode (num_tokens <= 64) must go through ... sparse_mla_sm120_decode_dsv4
+# (The main model uses topk=128 and is unaffected; only the DSpark draft asks for 256.)
+#
+# Fix: round an unsupported draft topk DOWN to the nearest supported bucket (256 -> 128)
+# right before dispatch — slice the indices to that width and clamp topk_length.
+#
+# Why this is correctness-safe: the DSpark draft is a SPECULATOR. Every token it proposes
+# is verified by the full target model before emission, so reshaping the draft's own
+# attention can only change accept-length (speed), never the emitted output. Rounding down
+# drops the lowest-ranked sparse candidates from the draft's attention only.
+#
+# If draft accept-length suffers, replace the "round down" with "pad up" (256 -> 512 with
+# -1 sentinels + real topk_length) to keep all candidates — lossless but needs matching
+# scratch (num_splits) sizing.
+set -e
+F="$(python3 -c 'import os,flashinfer; print(os.path.join(os.path.dirname(flashinfer.__file__),"mla","_sparse_mla_sm120.py"))' 2>/dev/null)"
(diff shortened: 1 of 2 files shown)


gitmarecki:
this PR helped me to fix Deepseek 0731 release! thank you!!!

The topk=256 diagnosis is spot on and saved me a ton of time. Confirmed working end-to-end on deepseek-ai/DeepSeek-V4-Flash-0731 (2x DGX Spark, TP=2, dspark method, 7 speculative tokens).

One bug in the patch as submitted, though: the str.replace anchor matches inside get_sparse_mla_sm120_module(), but model_type/topk_length are parameters of the nested _paged_attention closure — they don't exist in the outer scope. The anchor text technically matches (Python's in check doesn't care about indentation), so run.sh applies "successfully" and passes the AST check, but the injected block lands one indent level too shallow and ends up outside _paged_attention. Result: NameError: name 'model_type' is not defined at cache-build time instead of the topk fix ever running.

Fix is to anchor on topk = indices.shape[-1] inside _paged_attention itself (8-space indent, not 4) and inject there instead. Corrected run.sh:

```
#!/bin/bash
set -e
F="$(python3 -c 'import os,flashinfer; print(os.path.join(os.path.dirname(flashinfer.__file__),"mla","_sparse_mla_sm120.py"))' 2>/dev/null)"
[ -f "$F" ] || { echo "[fix-dspark-sm120-topk] ERROR: flashinfer _sparse_mla_sm120.py not found ($F)" >&2; exit 1; }

python3 - "$F" <<'PY'
import sys
f = sys.argv[1]
s = open(f).read()
MARKER = "fix-dspark-sm120-topk-v2"
if MARKER in s:
    print("[fix-dspark-sm120-topk] already applied"); raise SystemExit(0)

anchor = "        num_tokens, num_heads, d_qk = q.shape\n        topk = indices.shape[-1]\n"
if anchor not in s:
    raise SystemExit("[fix-dspark-sm120-topk] ERROR: anchor not found — flashinfer version drift")

inject = (
    "        # " + MARKER + ": round an unsupported topk down to the nearest\n"
    "        # bucket _decode_dsv4_dispatchable() supports. Draft-only — target\n"
    "        # verifies every proposal, so this affects accept-length, not output.\n"
    "        if model_type == _MODEL_TYPE_DSV4:\n"
    "            _DSV4_TOPK_OK = (128, 512, 1024)\n"
    "            if topk not in _DSV4_TOPK_OK:\n"
    "                _lo = [b for b in _DSV4_TOPK_OK if b <= topk]\n"
    "                _tgt = max(_lo) if _lo else min(_DSV4_TOPK_OK)\n"
    "                indices = indices[..., :_tgt].contiguous()\n"
    "                if topk_length is not None:\n"
    "                    topk_length = topk_length.clamp(max=_tgt)\n"
    "                topk = _tgt\n"
)
open(f, "w").write(s.replace(anchor, anchor + inject, 1))
print("[fix-dspark-sm120-topk] applied inside _paged_attention (v2)")
PY
python3 -c "import ast; ast.parse(open('$F').read()); print('[fix-dspark-sm120-topk] verified (AST valid)')"
```

maci0:
FlashInfer already landed TOPK 192 and 256 in `_DECODE_DSV4_DISPATCH` (flashinfer-ai/flashinfer#4380, merged 2026-08-08). Rounding DSpark draft 256 down to 128 is obsolete on current flashinfer and costs accept length.

For `deepseek-ai/DeepSeek-V4-Flash-0731`, k=5 with window 128 needs **192**, not 128: `ceil((128+5)/64)*64 = 192`. Adding the `(H, 192)` dispatch entries (Python + `sparse_mla_sm120_decode_dsv4.cu`) is what we run on 2x GB10. Mean accept was ~2.0-3.5, greedy France stayed `" Paris. The capital of Spain is Madrid..."`.

This recipe pins `DeepSeek-V4-Flash-DSpark`. The 0731 checkpoint is a different pin (`dspark_block_size=5` in the 0731 config) and already has `recipes/deepseek-v4-flash-0731.yaml` on main.

Issue #230, opened 2026-05-05
Build cache?


uacode:
_I created a recipe and changed the model name, but it turns out it's being cached somehow?_
> My facap, the model changed only in the name at the top

And I don't understand why I have different versions of vLLM for different models?

```
root@spark-1:/opt/spark-vllm-docker# ./run-recipe.sh qwen3.6-35b-a3b-nvfp4 --solo
Recipe: Qwen3.6-35B
  vLLM serving Qwen3.6-35B-A3B-NVFP4 solo on GB10 (128 GB unified memory)

=== Launching ===
Container: vllm-node-tf5
Mode: Solo

Loading configuration from .env file...
Loaded .env variables: DOTENV_CONTAINER_HF_TOKEN DOTENV_CONTAINER_NAME DOTENV_CONTAINER_NCCL_DEBUG DOTENV_CONTAINER_NCCL_IGNORE_CPU_AFFINITY DOTENV_MASTER_PORT 
Adding container env: HF_TOKEN
Adding container env: NCCL_DEBUG
Adding container env: NCCL_IGNORE_CPU_AFFINITY
Using launch script: /tmp/tmpieqgg74u.sh
Solo mode enabled. Skipping node detection.
Head Node: 127.0.0.1
Worker Nodes: 
Container Name: vllm_node
Image Name: vllm-node-tf5
Action: exec
Starting Head Node on 127.0.0.1...
48a77bd684bd485ade89cc1f41565980eef53144b8e4692eac91a006a0f77e56
Copying launch script to head node...
Successfully copied 2.56kB to vllm_node:/workspace/exec-script.sh
Executing command: /workspace/exec-script.sh
(APIServer pid=41) INFO 05-05 18:27:27 [utils.py:299] 
(APIServer pid=41) INFO 05-05 18:27:27 [utils.py:299]        █     █     █▄   ▄█
(APIServer pid=41) INFO 05-05 18:27:27 [utils.py:299]  ▄▄ ▄█ █     █     █ ▀▄▀ █  version 0.19.1rc1.dev374+g1174723eb.d20260417
(APIServer pid=41) INFO 05-05 18:27:27 [utils.py:299]   █▄█▀ █     █     █     █  model   RedHatAI/Qwen3.6-35B-A3B-NVFP4
(APIServer pid=41) INFO 05-05 18:27:27 [utils.py:299]    ▀▀  ▀▀▀▀▀ ▀▀▀▀▀ ▀     ▀
[...]

Issue #228, opened 2026-05-04
Patching vllm for Qwen3-Coder-Next doesn't work


0k00l:
Patching vllm for Qwen3-Coder-Next doesn't work. Version of spark-vllm-docker is up to date (cloned today 2026-05-04)

**Patching Qwen3-Coder-Next crashing on start
patching file vllm/v1/core/single_type_kv_cache_manager.py
Hunk #1 FAILED at 1000.
1 out of 1 hunk FAILED -- saving rejects to file vllm/v1/core/single_type_kv_cache_manager.py.rej**

```bash
ascent@ascent:~/spark-vllm-docker$ ./run-recipe.sh qwen3-coder-next-fp8 --solo --port 8000 --host 0.0.0.0
Recipe: Qwen3-Coder-Next-FP8
  vLLM serving Qwen3-Coder-Next-FP8
=== Launching ===
Container: vllm-node
Mods: mods/fix-qwen3-coder-next
Mode: Solo

Using launch script: /tmp/tmp688gps6o.sh
Solo mode enabled. Skipping node detection.
Head Node: 127.0.0.1
Worker Nodes: 
Container Name: vllm_node
Image Name: vllm-node
Action: exec
Starting Head Node on 127.0.0.1...
0ec3747d295b330ff2f3cb1a85b6c01289c013a257d45f23aafbb1f19a6c8a99
Applying modifications to cluster nodes...
Applying mod 'fix-qwen3-coder-next' to 127.0.0.1...
  Copying directory content to container...
Successfully copied 9.73kB to vllm_node:/workspace/mods/fix-qwen3-coder-next/
  Running patch script on 127.0.0.1...
Patching Qwen3-Coder-Next crashing on start
patching file vllm/v1/core/single_type_kv_cache_manager.py
Hunk #1 FAILED at 1000.
1 out of 1 hunk FAILED -- saving rejects to file vllm/v1/core/single_type_kv_cache_manager.py.rej
Patch is not applicable, skipping
Reverting PR #34279 that causes slowness
patching file vllm/model_executor/layers/fused_moe/fused_moe.py
Unreversed patch detected!  Ignore -R? [n] 
Apply anyway? [n] 
Skipping patch.
2 out of 2 hunks ignored -- saving rejects to file vllm/model_executor/layers/fused_moe/fused_moe.py.rej
Can't revert PR #34279, skipping as it was reverted in recent commits
Fixing Triton allocator bug
Copying launch script to head node...
[...]

Issue #214, opened 2026-04-27
instantensor and tool calling in gemma4-26b-a4b recipe


vincenzo-antolini-os:
Recipe Gemma4 26B A4B.

`instanttensor `makes it broken, loops and weird random hangs.
I changed to `safetensor`.
Didn't try `fasttensors`.

Honestly, it was not so easy to spot 'cause error happens after some requests.

Latest repo build get vLLM 0.19.2rc (if I remember correctly) and I think gemma4 mod has been merged.

```
Applying mod 'fix-gemma4-tool-parser' to 127.0.0.1...
  Copying directory content to container...
Successfully copied 2.56kB to vllm_node:/workspace/mods/fix-gemma4-tool-parser/
  Running patch script on 127.0.0.1...
Applying PR #38909
error: patch failed: vllm/tool_parsers/gemma4_tool_parser.py:436
error: vllm/tool_parsers/gemma4_tool_parser.py: patch does not apply
- PR #38909 can't be applied, skipping
```

thewildcat:
Did you try using the chat template thats including in llama.cpp? https://github.com/ggml-org/llama.cpp/blob/master/models/templates/google-gemma-4-31B-it.jinja this works fine for me even with instanttensor loading.

vincenzo-antolini-os:
Hi, no... I didnt, cause I'm a little scared messing with vLLM.
I changed to `instanttensor` some days ago and it made a bad first impression to me. Broken tools after some requests.

As far as I understood instanttensor/safetensors is  affecting only loading time, not "inference" time.
Is it correct?

PR #324, opened 2026-07-23
Fix JSON-valued CLI args that sparkrun's renderer cannot substitute

brainchillz:
## Bug

Running `deepseek-v4-flash` (and 8 other recipes) with current sparkrun (0.2.40) fails at vllm argument parsing:

```
usage: vllm serve [model_tag] [options]
vllm serve: error: argument --speculative-config/-sc: Value {{"method":"mtp","num_speculative_tokens":{num_speculative_tokens}}} cannot be converted to <function loads at 0xecd492b294e0>.
```

The template text reaches vllm completely un-substituted.

## Root cause

The recipes use `str.format()`-style conventions (`{{` brace escaping, placeholders nested inside JSON), but sparkrun renders `command:` templates with a regex substituter (`vpd.legacy.arguments.arg_substitute`, non-greedy `\{(.*?)\}`), called in a loop from `Recipe.render_command`. Two consequences:

1. `{{` is not an escape — it is just a literal character the regex trips over.
2. For a token like `'{{"method":"mtp","num_speculative_tokens":{num_speculative_tokens}}}'`, the first regex match spans from the leading brace to the *first* `}` — the extracted "variable name" is the whole JSON blob, which matches nothing, so the original text is put back verbatim. The inner placeholder is unreachable no matter how many passes run.

Even JSON args with no inner placeholder (e.g. `--diffusion-config '{{"canvas_length":256}}'`) are broken: the doubled braces are never collapsed, so vllm receives `{{...}}`, which is invalid JSON.

## Fix

Move each JSON blob into a `defaults:` variable and reference it whole:

```yaml
defaults:
  speculative_config: '{"method":"mtp","num_speculative_tokens":2}'
command: |
  ... --speculative-config '{speculative_config}' \
```

`render_command`'s nested-substitution loop resolves the variable on the first pass; on the second pass the substituted single-brace JSON matches no known variable and passes through untouched. As a bonus the JSON stays overridable via user config, like any other default.

`deepseek-v4-flash`'s `num_speculative_tokens` default is folded into `speculative_config` (it could never be substituted independently anyway).

## Scope

18 args across 9 recipes: `deepseek-v4-flash`, `qwen3.6-35b-a3b-nvfp4`, `qwen3.6-35b-a3b-fp8-dflash`, `gemma4-26b-a4b-nvfp4`, `4x-spark-cluster/nemotron-3-ultra-nvfp4`, and the four `diffusion-gemma-*` variants.

## Verification

- All 31 recipes in the repo rendered through sparkrun 0.2.40's `Recipe.render_command`; every `--*-config` / `--default-chat-template-kwargs` value now parses with `json.loads`, and no `{placeholder}` tokens remain.
- End-to-end on a 2-node DGX Spark cluster: the fixed `deepseek-v4-flash` recipe launches, vllm passes argparse with `--speculative-config {"method":"mtp","num_speculative_tokens":2}`, serves `/health`, and completes chat requests.

### DIFF
--- a/recipes/4x-spark-cluster/nemotron-3-ultra-nvfp4.yaml
+++ b/recipes/4x-spark-cluster/nemotron-3-ultra-nvfp4.yaml
@@ -28,6 +28,8 @@ defaults:
   max_model_len: 262144
   max_num_seqs: 2
   max_num_batched_tokens: 16384
+  # JSON-valued args live here whole: the renderer can't substitute inside JSON braces
+  speculative_config: '{"method":"mtp","num_speculative_tokens":4,"moe_backend":"triton","attention_backend":"TRITON_ATTN"}'
 
 command: |
   vllm serve nvidia/NVIDIA-Nemotron-3-Ultra-550B-A55B-NVFP4 \
@@ -54,7 +56,7 @@ command: |
     -cc.mode 0 \
     -cc.cudagraph_mode FULL_DECODE_ONLY \
     --cudagraph-capture-sizes 5 10 15 20 \
-    --speculative-config '{{"method":"mtp","num_speculative_tokens":4,"moe_backend":"triton","attention_backend":"TRITON_ATTN"}}' \
+    --speculative-config '{speculative_config}' \
     --enable-prefix-caching \
     --enable-flashinfer-autotune \
     --reasoning-parser nemotron_v3 \
--- a/recipes/deepseek-v4-flash.yaml
+++ b/recipes/deepseek-v4-flash.yaml
@@ -27,7 +27,9 @@ defaults:
   block_size: 256
   max_num_seqs: 4
   max_num_batched_tokens: 8192
-  num_speculative_tokens: 2
+  # JSON-valued args live here whole: the renderer can't substitute inside JSON braces
+  speculative_config: '{"method":"mtp","num_speculative_tokens":2}'
+  reasoning_config: '{"reasoning_parser":"deepseek_v4","reasoning_start_str":"","reasoning_end_str":""}'
 
 # Environment variables
 env:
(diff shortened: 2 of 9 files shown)


brainchillz:
This seriously bit me twice ... I fixed it by hand once and then later ran a spark run update and it undid my manual fix by reloading these broken recipes over top of it.

eugr:
@dbotwinick - I believe you added compatibility with this recently?

brainchillz:
It doesn't look like any of the recipes have been fixed in eugr/spark-vllm-docker. Main moved from 4851a09 to 81a33f3 (3 commits: flashinfer cleanup, mod docs, a mamba/MTP mod), and all 9 recipes still ship the broken {{...}}
  templates. I verified end-to-end: rendering the current upstream deepseek-v4-flash.yaml through the installed sparkrun 0.2.40 still passes the raw
  '{{"method":"mtp","num_speculative_tokens":{num_speculative_tokens}}}' template straight to vllm.

  The "compatibility" you mentioned to dbotwick is real but appears unreleased and incomplete. In spark-arena/sparkrun (https://github.com/spark-arena/sparkrun), PR #214 (https://github.com/spark-arena/sparkrun/pull/214)
  merged a fix into develop-03x (the 0.3.0 alpha line) that collapses {{ → { in render_command after substitution, and dbotwinick closed PR #230 (https://github.com/spark-arena/sparkrun/pull/230) and issue
  #228 (https://github.com/spark-arena/sparkrun/issues/228) as "already resolved in 0.3.0 alpha". However:

  1. Nothing is released — PyPI latest is still 0.2.40 and neither main nor develop (0.2.41-dev) has the fix.
  2. The alpha fix misses nested placeholders. I simulated the algorithm (substitution loop, then brace collapse): the literal-JSON recipes like diffusion-gemma render correctly, but
  deepseek-v4-flash still produces {"method":"mtp","num_speculative_tokens":{num_speculative_tokens}} — invalid JSON, because the non-greedy regex consumes the inner placeholder before it can substitute.
  So even on 0.3.0 alpha, that recipe stays broken; this defaults-variable approach in PR #324 is still the only thing that fixes it.

brainchillz:
It's not hurting me anymore, I have durable local copies that have been fixed in my machines that aren't being overwritten by the munged ones being pushed in the updates anymore but it's got to be getting in someone elses way out there somewhere. :D

PR #329, opened 2026-07-31
Add Inkling recipe with tool-call parser fix mods

IMJONEZZ:
Inkling-small-nvfp4 recipe, first off.
If you run this as-is, you'll get tool-calling unparsed due to inkling generating " and &quot interchangeably, which vLLM does NOT appreciate. It also has a problem with multi-turn tool-calling, which this addresses as well.

- inkling-tool-unescape: the model occasionally samples an HTML entity (usually &quot;) in place of a structural JSON quote, most often the closing quote right after a URL. This breaks the Inkling tool parser's hand-rolled string scanner, so the tool call fails and the raw <|content_invoke_tool_json|> block leaks as content. The mod unescapes complete JSON-structural entities in _inkling_arg_converter before the scan; prefix-stable for streaming.

- inkling-tool-salvage: in multi-turn streaming, a turn-initial tool call leaks as content with tool_calls empty because the streaming DelegatingParser drops the content-kind marker that starts the tool block (deterministic: multi-turn + streaming + the assistant turn starting directly with a tool call; single-turn and non-streaming are fine, independent of reasoning_effort). The mod wraps DelegatingParser.parse_delta, buffers streamed content, and re-emits complete <|content_invoke_tool_json|>{...}<|end_message|> blocks as tool_calls; it also strips a leaked trailing <|end_message|>. Gated by the Inkling marker string, so it is a no-op for other models.

Both are idempotent and refuse to patch if the target vLLM code has changed. They are downstream band-aids pending upstream vLLM fixes (vllm-project/vllm#50512, #49865); remove once those land.

I will continue working on this, but this makes inkling-small-nvfp4 at least usable on a 2-spark cluster, any help or advice on how to get it more stable would be appreciated but I fully expect both a re-release of the inkling-small chat template and a vLLM inkling tool parser patch in the next week or so to address it, at which point the recipe will work without the mods.


AndreasKunar:
Thanks a lot, your PR worked for me, so that I could use Inkling in openclaw!!!

The quality of Inkling Small NVFP4 as agentic LLM has not impressed me much. To me, it's very talkative and looping instead of progressing. DeepSeek V4 Flash 0731, which needs comparable resources (dual Spark), is much, much better as LLM. 

But I still need to see how image/voice work, especially voice internationally. And which performance I get for its multimodality. Its real-time conversation capabilities should be the main advantage.

PR #332, opened 2026-08-01
Support building vLLM from a fork (VLLM_REPO, SETUPTOOLS_SCM_PRETEND_VERSION)

tobymao:
The vLLM remote was hardcoded to vllm-project/vllm, so the image could only be
built from upstream. Hardware-enablement work often lives on an unmerged fork
branch -- for example SM121/GB10 DeepSeek-V4 support in vllm-project/vllm#41834 --
and those users currently have to patch the Dockerfile.

Two additions, both defaulting to current behaviour so stock builds are unchanged:

1. ARG VLLM_REPO and --vllm-repo. The cache-hit path also repoints the remote with
   git remote set-url, otherwise a cached upstream clone would silently keep
   fetching from the wrong repository.

2. ARG SETUPTOOLS_SCM_PRETEND_VERSION and --scm-version. setuptools_scm derives the
   wheel version from the nearest git tag, and fork branches often carry tags that
   are not valid PEP 440 -- a respin suffixed "-20260727d" yields the candidate
   version "20260727d" and the build fails before compiling anything. Applied
   inline on the build step via :+ so an unset value is a true no-op.

Tested by building DeepSeek-V4-Flash-0731 from jasl/vllm@d64074e6 on 4x DGX Spark
(GB10/SM121); the resulting image serves and passes a correctness probe.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -242,6 +242,9 @@ WORKDIR $VLLM_BASE_DIR
 # --- VLLM SOURCE CACHE BUSTER ---
 ARG CACHEBUST_VLLM=1
 
+# Upstream repository to build from. Parameterized so the image can be built from a
+# fork -- e.g. hardware-enablement branches that are not merged upstream yet.
+ARG VLLM_REPO=https://github.com/vllm-project/vllm.git
 # Git reference (branch, tag, or SHA) to checkout
 ARG VLLM_REF=main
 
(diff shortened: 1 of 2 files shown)


eugr:
`--vllm-repo` is now supported as an argument to `./build-and-copy.sh`

PR #333, opened 2026-08-02
fix: broken speculative-config JSON in deepseek-v4-flash recipe

dplundberg:
## Bug

The `deepseek-v4-flash` recipe (v1 format) produces invalid JSON for `--speculative-config`, causing vLLM to fail on launch:

```
vllm serve: error: argument --speculative-config/-sc: Value
{"method":"mtp","num_speculative_tokens":{num_speculative_tokens}}
cannot be converted to <function loads>.
```

## Root Cause

The v1 recipe escapes literal JSON braces as `{{` and `}}` to protect them from `{placeholder}` substitution. However, sparkrun's `arg_substitute` uses regex `\\{(.*?)\\}` (non-greedy), which matches from the first `{` in `{{` to the first `}`. With doubled braces, it consumes the entire JSON string — including the inner `{num_speculative_tokens}` placeholder — as one giant unmatched key. The placeholder is never substituted. The post-substitution `{{ → {` collapse then produces invalid JSON with a bare `{num_speculative_tokens}`.

## Fix

Move both JSON config values (`speculative_config` and `reasoning_config`) into the recipe's `defaults` dict as complete string values, then reference them as simple `{speculative_config}` and `{reasoning_config}` placeholders in the command template. This matches the pattern already used by v2 recipes (e.g. `deepseek4-flash-fp8-mtp-vllm`).

### DIFF
--- a/recipes/deepseek-v4-flash.yaml
+++ b/recipes/deepseek-v4-flash.yaml
@@ -28,6 +28,8 @@ defaults:
   max_num_seqs: 4
   max_num_batched_tokens: 8192
   num_speculative_tokens: 2
+  speculative_config: '{"method":"mtp","num_speculative_tokens":2}'
+  reasoning_config: '{"reasoning_parser":"deepseek_v4","reasoning_start_str":"","reasoning_end_str":""}'
 
 # Environment variables
 env:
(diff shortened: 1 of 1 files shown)


mountaintopsolutions:
Fixes are being done in sparkrun to fix this without having to move to defaults.  See https://github.com/spark-arena/sparkrun/commit/14b667e39a35292eda805ed3fc9e786c82848a3c

eugr:
closing as being addressed in Sparkrun

PR #360, opened 2026-08-25
feat(recipes): adopt fleet scheduling for the ds4f 1M recipe

paulclou:
## Summary

Under a 12-agent coding-agent fan-out (44–90k-token contexts, every session resubmitting its grown prompt each turn), the 1M recipe's scheduling collapsed into a prefix-cache eviction cycle: 8 running + 4 queued for hours, cache hit rate 94% lifetime → 12% in the final 80-minute window, ~16.5M tokens re-prefilled in that window (~3,500 tok/s continuously) against 75k generated (~16 tok/s aggregate), queue p90 77.8s, one agent 75 minutes between completed turns. Zero preemptions throughout — the pathology is LRU eviction of queued agents' cached prefixes, not pool overflow.

Fix is to stop queueing the fleet at the server. Minimal change, one variable-pair, so the performance differential is attributable:

- `max_num_seqs` 8 → 12 — all concurrent working sets stay pinned (12 × ~85k ≈ 1.0M < 1.70M pool); batch decode also amortizes the per-step weight read.
- `max_cudagraph_capture_size` 64 → 128 — not independent: dspark k=5 makes a decode step carry ~6 tokens/seq, so 12 seqs = 72 tokens overflows 64 and decode falls off the captured graphs exactly when the new slots are used.

Deliberately **not** adopted: `max_num_batched_tokens` 4096 (the pool lever, measured −8% decode) — deferred until contexts grow past ~100k average or the fleet exceeds 12; the header documents the sizing rule.

The baseline pathology numbers are recorded in the recipe header's new FLEET SCHEDULING section so the post-change measurement can be appended beside them; boot caveat (seqs 12 + capture 128 not yet A/B-booted at the 1M window; fallback is capture 64) documented there too.

## Verification

- `./tests/test_recipes.sh`: 65 pass / 1 fail — the failure (`README match: minimax-m2-awq`) pre-exists on `main` and is unrelated.
- Dry-run (`--config /dev/null -n <dummy>`) confirms the generated command carries `--max-num-seqs 12` and `--max-cudagraph-capture-size 128` with everything else unchanged.


paulclou:
Opened against the wrong repo — this is fork-internal tuning; re-filed on the fork.

PR #373, opened 2026-09-02
Add mods/fix-kv-offload-disk-tier: EAGLE store filter + multi-node promoted-row re-sync

yaro-tal:
Two bugs in `OffloadingConnector` + `TieringOffloadingSpec` with an `fs`
secondary tier. They're in one mod because neither fix is useful alone: apply
only the first and a multi-node cluster gets a working cache that corrupts;
apply only the second and there's nothing to correct, because the cache never
hits.

Found while running DeepSeek-V4-Flash-0731 across two DGX Sparks (TP=2, one GPU
per node). Verified against vLLM `e2666d9a65f41fc376607531453cbd57c4c71016`.

## 1. EAGLE/MTP groups can never certify a hit — the tier returns zero hits

The store path drops sliding-window chunks no lookup could reach, keeping only
the trailing `tail` chunks of each alignment segment. But
`_sliding_window_lookup` finds `tail` chunks and an *unverified* EAGLE group
then pops one (`num_hit_chunks -= 1`), so it needs `tail + 1` **consecutive**
chunks — which the store filter has guaranteed it can never see.

Because `_lookup()` ANDs the per-group results, that group's permanent zero
vetoes every other group as well. The tier reports a 0% hit rate while happily
writing hundreds of GB. Measured here: **502 GB stored, 0 bytes ever read back.**

Fix: also keep the segment-head chunk for eagle groups, so the retained set is
`{0} ∪ {acc-tail .. acc-1}` — a run of exactly `tail + 1` consecutive chunks
across the segment boundary. The pop lands on a chunk at `0 (mod acc)`, so the
resulting length stays a multiple of the full-attention chunk size.

## 2. Multi-node TP silently serves wrong KV

`SharedOffloadRegion` is a **per-node `/dev/shm` mmap**. On a single node the
scheduler-side region (`rank=None`) and every worker region (`rank=r`) are the
*same file* — `rank` only selects a slot within a row — so a promotion the tier
manager performs is visible to every worker for free. That assumption is
undocumented, and false as soon as TP spans nodes: each node has its own mmap,
and only the rank co-located with the manager runs a secondary tier at all.

- GPU→CPU stores stay symmetric — every rank writes its own region
- disk→CPU promotions land **only on the manager's rank**
- the following CPU→GPU load reads **each rank's own region**

So every other rank feeds its GPU whatever stale bytes occupied that row.
Nothing raises, nothing warns, no checksum fails.

Hashing the same row index in both nodes' mmaps:

| rows | identical across ranks (before) | after |
|---|---|---|
| written by GPU→CPU stores | 112 / 112 | 112 / 112 |
| written by disk→CPU promotion | **0 / 112** | **112 / 112** |

Over a longer run the correspondence was exact: every divergent row was a
promoted row, and every promoted row was divergent.

The symptom depends only on what was in the stale row, which is why it presents
as several unrelated bugs — zeros (fresh mmap) give coherent output with
confabulated later content; a recycled row gives token soup and **other
sessions' content bleeding into the response**. Hits served entirely from the
CPU tier are always correct, because no promotion is involved, which is what
makes it hard to catch.

Fix: the manager records the rows a *completed* promotion filled,
`build_connector_meta()` drains them into
`OffloadingConnectorMetadata.promoted_rows`, and every rank re-syncs them from
the manager's rank over the existing TP group **before any load is submitted**.
Ordering matters: `build_connector_meta()` runs `on_schedule_end()` →
completed-job processing first, so a promotion that lands in a step ships its
row ids in that same step's metadata. All ranks get an identical list, so they
issue the same collectives in the same order.

Read once and transfer over the link, rather than having every rank read the
tier — a second reader pulls the same bytes over the same link anyway *and*
hits the disk twice. **Consequence worth having: the tier no longer needs to be
shared between nodes**, so it can be node-local with no NFS.

### Safety gate

Re-syncing one rank's rows onto another is only correct for a **replicated** KV
cache. MLA stores one compressed latent per token and is replicated by
construction; a head-sharded cache (GQA/MHA) or per-rank recurrent state
genuinely differs per rank, and copying over it would corrupt it just as
thoroughly as the bug. So the mod does nothing unless every KV group is
known-replicated, and logs which way it decided.
`VLLM_OFFLOAD_KV_REPLICATED=0|1` overrides. A one-shot check immediately after
the first broadcast confirms the rows landed identically — race-free, since
equality there holds by construction.

**On single-node deployments patch 2 is a no-op**, so it's safe to leave applied.

## Testing

- Both patches apply in sequence to a pristine image, `compileall` clean,
  `import vllm` OK, and `run.sh` is idempotent on a second run.
- End-to-end on the two-node cluster: store a prompt, evict it with five filler
  prompts, load it back from disk. Before: multilingual token soup. After: all
  checkpoints correct, with the row diff going 0/112 → 112/112 identical.
- Verified the load really came from disk (112 `fs` reads), not a CPU-tier hit —
  `reset_prefix_cache` does *not* drain the CPU tier at `blocks_per_chunk > 1`,
  which makes naive tests measure memory instead.

## Notes

`PYTHONHASHSEED` must be set to the same fixed value everywhere or `NONE_HASH`
is reseeded per process and nothing on disk is ever found again. vLLM already
warns about this; the tier makes it expensive. Worth knowing when checking:
`/proc/<pid>/environ` is unreliable for `VLLM::EngineCore`, which calls
`setproctitle` and clobbers that region — it cost us a wrong diagnosis.

Happy to split this into two mods if you'd rather, though they only produce a
working tier together.

### DIFF
--- a/mods/fix-kv-offload-disk-tier/01-eagle-store-filter.patch
+++ b/mods/fix-kv-offload-disk-tier/01-eagle-store-filter.patch
@@ -1,0 +1,37 @@
+diff --git a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py
+index 0d15ec2..482cff6 100644
+--- a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py
++++ b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py
+@@ -995,6 +995,10 @@ class OffloadingConnectorScheduler:
+ 
+                 alignment_chunk_count = group_config.alignment_chunk_count
+                 tail = group_config.sliding_window_size_in_chunks
++                # The EAGLE certificate chunk sits at the segment boundary.
++                eagle_keep_segment_head = (
++                    tail is not None and group_config.is_eagle_group
++                )
+ 
+                 for key_idx, (offload_key, block_id) in enumerate(
+                     zip(offload_keys, offload_block_ids)
+@@ -1010,7 +1014,20 @@ class OffloadingConnectorScheduler:
+                         assert tail is not None
+                         abs_chunk_idx = start_chunk_idx + key_idx
+                         pos_in_segment = abs_chunk_idx % alignment_chunk_count
+-                        if pos_in_segment < alignment_chunk_count - tail:
++                        keep = pos_in_segment >= alignment_chunk_count - tail
++                        # EAGLE/MTP groups additionally need the chunk at the
++                        # segment head. _sliding_window_lookup finds `tail`
++                        # chunks, but an unverified eagle group then pops one
++                        # (num_hit_chunks -= 1), so it must see tail + 1
++                        # CONSECUTIVE chunks or it certifies nothing. Keeping
++                        # {0} u {acc-tail .. acc-1} forms exactly such a run
++                        # across the segment boundary. Because _lookup() ANDs
++                        # the per-group results, an eagle group that can never
++                        # certify vetoes every other group as well -- which is
++                        # why the tier returns zero hits without this.
++                        if eagle_keep_segment_head and pos_in_segment == 0:
++                            keep = True
++                        if not keep:
+                             continue
+                     new_offload_keys.append(offload_key)
+ 
--- a/mods/fix-kv-offload-disk-tier/02-multinode-promoted-row-resync.patch
+++ b/mods/fix-kv-offload-disk-tier/02-multinode-promoted-row-resync.patch
@@ -1,0 +1,361 @@
+diff --git a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/common.py b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/common.py
+index 939a2b0..a5449c5 100644
+--- a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/common.py
++++ b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/common.py
+@@ -71,6 +71,11 @@ class OffloadingConnectorMetadata(KVConnectorMetadata):
+     store_jobs: dict[int, TransferJob]
+     jobs_to_flush: set[int] | None = None
+ 
++    # Primary-tier block ids filled by a secondary-tier promotion during this
++    # step. Workers that do not share the tier manager's mmap must re-sync
++    # these before any load reads them; see OffloadingConnectorWorker.
++    promoted_rows: list[int] | None = None
++
+ 
+ @dataclass
+ class OffloadingWorkerMetadata(KVConnectorWorkerMetadata):
+diff --git a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py
+index 482cff6..6694897 100644
+--- a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py
++++ b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/scheduler.py
+@@ -1188,10 +1188,19 @@ class OffloadingConnectorScheduler:
+                 for jid in self._block_id_to_pending_jobs[bid]
+             )
+ 
++        # on_schedule_end() above already ran the tier manager's completed-job
++        # processing, so promotions that landed this step travel in the SAME
++        # metadata as the load job that will read them.
++        promoted_rows = None
++        drain = getattr(self.manager, "drain_promoted_rows", None)
++        if drain is not None:
++            promoted_rows = drain() or None
++
+         meta = OffloadingConnectorMetadata(
+             load_jobs=self._current_batch_load_jobs,
+             store_jobs=self._build_store_jobs(scheduler_output),
+             jobs_to_flush=self._current_batch_jobs_to_flush,
++            promoted_rows=promoted_rows,
+         )
+         self._current_batch_load_jobs = {}
+         self._current_batch_jobs_to_flush = set()
+diff --git a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/worker.py b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/worker.py
+index 045e513..46ce1ad 100644
+--- a/vllm/distributed/kv_transfer/kv_connector/v1/offloading/worker.py
++++ b/vllm/distributed/kv_transfer/kv_connector/v1/offloading/worker.py
+@@ -1,5 +1,6 @@
+ # SPDX-License-Identifier: Apache-2.0
+ # SPDX-FileCopyrightText: Copyright contributors to the vLLM project
++import os
+ from collections import defaultdict
+ from dataclasses import replace
+ 
+@@ -33,6 +34,103 @@ from vllm.v1.kv_offload.base import (
+ 
+ logger = init_logger(__name__)
+ 
++# ---------------------------------------------------------------------------
++# Multi-node re-sync of promoted primary-tier rows.
++#
++# SharedOffloadRegion is a PER-NODE /dev/shm mmap. On a single node the
++# scheduler-side region (rank=None) and every worker region (rank=r) are the
++# SAME FILE -- rank only selects a slot within a row -- so a promotion the tier
++# manager performs is visible to every worker for free. That assumption is
++# undocumented, and it is false as soon as TP spans more than one node: each
++# node has its own mmap, and only the rank co-located with the manager runs a
++# secondary tier at all.
++#
++# The result is silent KV corruption. GPU->CPU stores stay symmetric (every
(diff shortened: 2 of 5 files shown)


yaro-tal:
## Status update: this mod is incomplete for the case it targets, and I know why now

Six days of production data later, I want to be straight about a limitation rather than leave it for someone to discover.

**The two patches here make the disk tier *correct*. They do not make it *useful* on a prefix larger than your primary tier.** If you apply this mod and your prefixes are long, you will most likely see the tier do nothing at all — no error, no warning, just persistent zero hits. That is not this mod failing to apply; it is a third bug underneath it.

### The bug

`TieringOffloadingManager.lookup` ends with:

```python
return LookupResult.MISS if not promoted else LookupResult.RETRY
```

The matching walk promotes **one primary-tier row per queried key**, just to confirm the key is there. Once the primary tier is full, every subsequent key — *including keys that are sitting on disk* — comes back `MISS`, indistinguishable from "never stored". The cross-group AND in the scheduler (`if num_hit_chunks == 0: return 0`) then discards the entire external hit.

It is self-reinforcing: the walk eats the same rows `prepare_store` needs, so the tier stops being **written** too, and it never recovers on its own.

### Why this will hit essentially everyone using this image

The primary tier is sized out of host RAM, and we are all on GB10 boxes with **128 GB total**. With a large model resident there is very little left — our `cpu_bytes_to_use: 4294967296` buys exactly **498 rows**. A long agent conversation queries far more keys than that during a single match. So the ceiling is not an exotic configuration, it is the default situation for this hardware. If you have 128 GB and long prefixes, you are in it.

### How to tell in 30 seconds

On a vetoed lookup, **sum the per-group RETRY counts**. If the sum pins at exactly your primary tier's row count, you are hitting this and your data is on disk, not missing.

You can confirm the row count independently from metric granularity: `kv_offload_cpu_cache_usage_perc` only ever takes values `k/rows` (ours reported `0.08032128514056225` = exactly `40/498`).

Do **not** read a high miss count as disk absence without doing this first. That misreading cost us about two weeks.

### Work in progress

The fix is separating *matching* from *staging* — match with `promote=False`, then stage the confirmed hit once. It is running in production here now:

| | before | after |
|---|---|---|
| summed RETRY per probe | 498 | **0** |
| `exit=ZERO` | 14 | **0** |
| `cannot store chunks` | 2 | **0** |
| cold 294,186-token prompt | `ext=0` | **`ext=290,816`** (98.9%) |

Across 40 requests: 88.3% of prompt tokens covered, **1,349,632 tokens not re-prefilled**.

I would rather land that here as a third patch than leave the mod half-useful, so **please treat this PR as not-ready-to-merge for now** unless you want the correctness fixes on their own. I will follow up with the patch. Two things are still unverified and I will not claim otherwise: the second rank's GPU-side KV is confirmed only at the CPU-row mirror, and the multi-wave path has never actually executed (see below).

A second-order note for anyone tuning: with the ceiling gone, the remaining pressure is **concurrency**. One 440k-token prefix needs ~222 of our 498 rows, so two concurrent large requests fit and three do not. An admission gate is written but deliberately switched off until it has been exercised.

### Related

- Companion PR eugr/spark-vllm-docker#381 — an unrelated but adjacent prefix-cache defect: templateless models get their Anthropic inline `system` messages hoisted to offset 0, which cost us the *entire* prefix cache on every Claude Code turn (~78% → 98.3% hit rate). Different layer, same symptom class of "silently re-prefills everything".
- Upstream vllm-project/vllm#53569 ("fs tier gets zero hits on multi-group DSV4F") may well be this same bug seen from the outside — I have inferred that from the title and have not yet read the issue body, so treat it as a lead, not a finding.
- Upstream RFC vllm-project/vllm#54363 describes the wrong-bytes failure class that patch `02` here addresses.

### Provenance

Written with AI assistance and verified on a live two-node cluster. These patches are a collaboration between Claude Opus 5 and DeepSeek-V4-Flash, each reviewing the other's work and keeping the other honest; every number above is measured from a real deployment rather than asserted by a model. Flagging it so any policy you have on AI-assisted contributions is your call, not my assumption.

yaro-tal:
## Patch 03 is in, and this is ready for review now

Following up on my earlier comment asking you to hold this: the missing piece has landed as `03-match-without-staging.patch`, and everything I said was unverified is now measured. **Please treat the PR as ready.**

### What 03 does

Matching promoted one primary-tier row per queried key just to confirm the key existed. Once the tier filled, keys that were *sitting on disk* returned `MISS` — indistinguishable from absence — and the cross-group AND threw away the whole hit. Self-reinforcing, because the walk ate the rows `prepare_store` needed.

Fix: match with `promote=False`, then stage the confirmed hit in waves so a hit larger than the tier can still be served.

### Results, all measured on a live 2-node TP=2 cluster

| | before | after |
|---|---|---|
| summed RETRY per probe | 498 | **0** |
| `exit=ZERO` | 14 | **0** |
| cold 294,186-token prompt | `ext=0` | **`ext=290,816`** (98.9%) |

Since the earlier comment, four things I would not claim without evidence:

- **Multi-wave genuinely runs** — 24 loads at `num_waves=2`/`3`. Slicing exact: `ext=305152 wave_sizes=[64, 64, 26]` is 64+64+21 = 149 g0 chunks = 305152/2048. Blocks close end to end, 512+512+190 = 1214 = 149x8 + 22.
- **Both ranks** — identical job ids and `src_blocks` on each; rank 1 emits the matching `cpu_to_gpu` transfers. The `src_offset`/`dst_offset` asserts survive wave boundaries, which is the pair that used to crash our second rank.
- **Streaming beats tier size, which is the whole point.** A **1 GiB / 124-row** tier served a cold **348,000-token** prompt at **99.5%**. That prefix needs ~170 rows to stage, so it is monolithically impossible — only waves can do it.
- **Output is not degraded.** Next-token distributions from tier-served KV sit *within the engine's own run-to-run noise floor*, and match vLLM's own GPU prefix cache (2.95 vs 2.93, floor 3.20). Worth knowing if you ever test this: the engine is **not** deterministic at `temperature=0`, so diffing generated text cannot distinguish corruption from noise — measure the noise floor first.

Zero asserts, zero aborts, zero `promote_refused` throughout.

### Things I want to be upfront about

- **Request parking ships OFF** (`VLLM_OFFLOAD_PARK=0`). It works — 20 slot events, every one released — but it is the least-exercised part and I would not default it on for other people.
- **The patch carries ~70 lines of env-gated diagnostics**, inert unless their env var is set. They are the instruments that found the bug; hand-removing them would mean shipping a file we have never run.
- **`cannot store chunks` is pre-existing and you may meet it**, because 03 makes small tiers useful enough that people will run small ones. Now quantified on our workload (~250-350k prompts): 1 GiB / 124 rows produced **8,870** warnings from 40 requests (worst one 1,847x, ~218/min); 2 GiB / 249 rows produced **0**. Raise `cpu_bytes_to_use` until it stops. Note this is *not* what `STREAM_WAVE_CHUNKS` controls — wave size governs the load path and is not referenced in the store path at all.
- `run.sh` also gained a whole-set idempotency guard: the per-patch reverse-check broke once a third overlapping patch existed, so a second run errored instead of skipping. Fixed and verified both ways.

Verified before pushing: applies cleanly to a from-scratch reconstruction of stock vLLM `e2666d9a6` + 01 + 02, every touched file compiles, `run.sh` runs the full stack end to end, and re-running skips.

### Provenance

Written with AI assistance and verified on the hardware described. A collaboration between Claude Opus 5 and DeepSeek-V4-Flash, each reviewing the other's work and keeping the other honest. Every number here is measured on a real deployment, not asserted by a model — including the ones that came out inconvenient.

yaro-tal:
### 11-hour soak update

The numbers in my previous comment came from short windows. Here they are after **11 hours of continuous real workload** (agent coding sessions, 250-350k-token prefixes), at `cpu_bytes_to_use` 2 GiB / 249 rows:

| | |
|---|---|
| requests | 153 |
| prompt tokens | 32,626,888 |
| **covered** | **90.1%** |
| **tokens not re-prefilled thanks to the tier** | **3,580,928** |
| cold arrivals (`local=0`) | 32 requests, 6.39M tokens, **54.6% served from disk** |
| multi-wave loads | 10 (4x `num_waves=2`, 6x `num_waves=3`) |
| parking slot events | 25, all released |
| `cannot store chunks` | **0** |
| asserts / tracebacks / `promote_refused` / aborts | **0** |

Two things this firms up specifically:

- The **2 GiB recommendation** in the README was based on a 55-minute window when I wrote it. Eleven hours later it is still zero, so the 1 GiB -> 2 GiB table can be read as a durable result rather than a snapshot.
- **Parking** has now taken and released the admission slot 25 times with no request ever failing to un-park, which was the failure mode I was most wary of. It still ships **off** by default.

Nothing in the patch or the README changes as a result — posting because the earlier evidence was thinner than I would want anyone to merge on.

NTillmann:
For anyone running this mod with wave streaming on (`VLLM_OFFLOAD_STREAM_WAVE_CHUNKS > 0`): we hit an engine crash in patch 03's wave driver after about 20 hours under load (`prepare_load` → `AssertionError: Block ... not found in cache`). The cause and a fix are proposed as patch 04 against this PR's branch: yaro-tal/spark-vllm-docker#1.

PR #374, opened 2026-09-03
refactor(serving): serve plain HTTP; TLS moves to tailscale serve

paulclou:
## Summary

**Problem.** PR #2 made the launcher inject `--ssl-certfile`/`--ssl-keyfile` into every `vllm serve` command. The certificate it relied on came from the retired pfSense CA, was issued for the dead `.leviathan` suffix, and vLLM only reads cert files at startup, so nothing could renew it. Every client saw a warning or needed a custom trust bundle.

**Solution.** Remove the injection and the `SSL_CERTFILE`/`SSL_KEYFILE` settings. TLS terminates in `tailscale serve` on the head node, which holds a publicly trusted certificate for `<node>.<tailnet>.ts.net` and renews it itself. Docs, the benchmarking guide, the ds4f recipe header, and the refusal-probe usage example move to the tailnet URL and drop the pfSense bundle steps.

**Scope.** Launcher code and docs only. No recipe command changes. Operational rollout (systemd drop-in with `--host 127.0.0.1`, `tailscale serve` on the head node, deleting the old cert files) happens outside this PR.

BREAKING CHANGE: the launcher no longer serves HTTPS by default. Clients that hit `https://<host>:8000` directly must switch to the tailnet URL, or pass the SSL flags after `--`.

## Verification

- `tests/test_recipes.sh`: 65 pass, 1 fail (`README match: minimax-m2-awq`), identical to origin/main.
- `tests/test_launch_cluster_image_sync.sh`, `tests/test_launch_cluster_vllm_pr.sh`: pass.
- Dry-runs of `glm-5.3-flash-uncensored-nvfp4` and `deepseek-v4-flash-0731-1m` (4 dummy nodes, `--config /dev/null`) and `qwen3.5-35b-a3b-fp8 --solo` render no `--ssl-*` flags.

Commit is unsigned (GPG agent was locked); can be re-signed while the PR is draft.


paulclou:
Opened against the wrong repository by mistake; this belongs on the paulclou fork. Closing, apologies for the noise.

PR #408, opened 2026-09-19
hf-download.sh: fix --copy-to silently shipping zero weights on current hf cache layout

Sggin1:
Two independent fixes to `hf-download.sh`, in separate commits so they can be taken independently.

---

## 1. `--copy-to` silently ships zero weights (the important one)

`copy_model_to_host()` uses `rsync -av` on the model directory. Current `hf` versions keep blob content in a **hub-level** store (`$HF_HOME/hub/blobs/<xx>/<hash>`) and make the model directory a symlink farm pointing at it. `rsync -a` preserves those as symlinks, so the peer node receives links whose targets were never transferred.

**The copy reports success.** On a dual-Spark TP=2 setup the worker ended up with 13 MB and 98 broken symlinks — 0 of 48 safetensors shards resolvable — while the script printed:

```
sent 12,275,744 bytes  received 1,170 bytes
total size is 12,261,112  speedup is 1.00
Copy to <peer-node> completed in 00:00:01
Copy complete.
```

The failure then surfaces much later as a vLLM worker startup error, far from the cause.

Models downloaded **before** the layout change have real files in the model dir and are unaffected, so existing caches keep working and only new downloads break. Since `hf-download.sh` runs `uvx hf download`, which resolves the newest `hf` on every invocation, every new download is exposed.

### Why the obvious fixes don't work

Measured against a real 156 GiB model, rsync 3.2.7:

| option | file list | total size | result |
|---|---|---|---|
| `-a` (current) | 190 (reg 43, link 133) | **12,261,112 B** | ships dangling links |
| `-a --copy-unsafe-links` | 190 (reg 43, link 133) | **12,261,112 B** | **no effect** — rsync doesn't classify these as unsafe |
| `-a --copy-links` (`-L`) | 190 (reg 176) | **335,663,717,228 B** | **2×** |

`-L` doubles because both link layers resolve independently: `snapshots/<sha>/*.safetensors -> ../../blobs/<hash>` **and** `blobs/<hash> -> ../../blobs/<xx>/<hash>`.

### This PR

Ships the model dir as-is (symlinks intact), then sends the referenced hub-level blobs separately, preserving relative paths.

The `??/` in the find pattern is load-bearing: **both** link kinds begin `../../blobs/`, but snapshot links resolve *inside* the model dir while hub-level links escape it. Matching only the sharded form selects exactly the hub-level set.

Verified: **49 entries, 0 missing, 167,819,616,863 B (156.3 GiB)** — the weights exactly once. Old-layout models produce an empty list, so the step is a no-op and the change is backward compatible.

> Note for anyone debugging this: `du -sh` on the model dir reports ~13 MB **even when the copy is correct** under the new layout, so it can't distinguish success from failure. `find <model_dir> -xtype l | wc -l` (expect 0) and `ls -lL <snapshot>/model-00001-of-*.safetensors` can.

---

## 2. `uvx` not found over non-interactive SSH

`uvx` is commonly installed to `~/.local/bin`, which isn't on PATH for `ssh host 'cmd'`. Driving `run-recipe.sh --download-only` remotely fails with `Error: 'uvx' command not found` despite uvx being installed. This adds `~/.local/bin` before erroring out.

---

## Testing

- `bash -n hf-download.sh` clean
- Blob-sync path exercised against a real 156 GiB model (dry-run and file-list verification)
- Old-layout model: empty blob list, no behaviour change
- Diff is additions only (28 lines), no line-ending churn

Happy to adjust the approach — if you'd prefer explicit layout detection or a version check over the symlink-pattern heuristic, say the word and I'll rework it.

*(Unrelated: the `~/.cache/b12x` mount I was going to report is already fixed on `main` — thanks.)*


eugr:
Thanks for flagging it! 
There is an easier fix - just committed the change.

